Issue #15, opened 2017-04-30
Loading of the main page takes over 30 seconds in Moodle 3.2


reactor0138:
Once you click on the student quiz (in moodle 3.2) it take over 30 seconds to load the page even with only 1 or 2 questions in the bank. This puts students off from question creation and being able to quickly do quizzes using the filters

frankkoch:
The initial opening of a StudentQuiz-activity might take a little long. I'm currently running a class with 160 questions and hundreds of quiz-executions. The initial opening takes about 10 seconds. A re-opening performs much faster.

We want to improve performance in further releases.

eXpl0it3r:
Please keep in mind that performance can also highly depend on your underlying system. If you're running this on a production system, please ensure that you've configured your web and MySQL/PostgreSQL server for best performance. If the possibility exists, you may also want to ask your server operators to increase memory or processing speed.

StudentQuiz is database heavy, so that might be the first step to try an optimize. We'll be running more performance tests and hope to find and fix some potential bottlenecks.

nwp90:
Seems the problem is that the query being run is causing multiple sequential scans over the entire question attempts table. And the whole query is being run twice (load_questions being run twice).

```
                                                                               QUERY PLAN                                         
                                      
----------------------------------------------------------------------------------------------------------------------------------
--------------------------------------
 Sort  (cost=2914592.18..2914593.39 rows=487 width=386)
   Sort Key: q.qtype, q.name
   ->  Merge Right Join  (cost=2741238.19..2914570.44 rows=487 width=386)
         Merge Cond: (qa.questionid = q.id)
         ->  Group  (cost=2183342.13..2289883.41 rows=5327064 width=24)
               Group Key: qa.questionid, (count(*)), (count(*))
               ->  Sort  (cost=2183342.13..2196659.79 rows=5327064 width=24)
                     Sort Key: qa.questionid, (count(*)), (count(*))
                     ->  Hash Left Join  (cost=1107346.08..1369680.32 rows=5327064 width=24)
                           Hash Cond: (qa.questionid = mdl_question_attempts_1.questionid)
                           ->  Merge Right Join  (cost=557767.97..799858.49 rows=5327064 width=16)
                                 Merge Cond: (mdl_question_attempts.questionid = qa.questionid)
                                 ->  Sort  (cost=557767.54..557798.68 rows=12456 width=16)
                                       Sort Key: mdl_question_attempts.questionid
[...]

PR #18, opened 2017-05-25
Pass conditions to joins where needed, to fix speed issues.

nwp90:
It's pretty ugly, but was the least bad way I could think of to do this for now.

### DIFF
--- a/classes/condition/studentquiz_condition.php
+++ b/classes/condition/studentquiz_condition.php
@@ -37,6 +37,44 @@ defined('MOODLE_INTERNAL') || die();
  */
 class studentquiz_condition extends \core_question\bank\search\category_condition {
 
+    /* Due to fix_sql_params not accepting repeated use of named params,
+       we need to get unique names for params that will be used more than
+       once...
+
+       init() from parent class duplicated here as we can't call it directly
+       (private) :-P
+
+       where() overridden with call to init() followed by call to parent
+       where()...
+
+       params() always returns $this->params, which doesn't change between
+       calls to get_in_or_equal, so don't need to fix anything there.
+       Which is fortunate, as there'd be no way to keep where() and params()
+       in sync.
+    */
+
+    protected function init() {
+        global $DB;
+        if (!$this->category = $this->get_current_category($this->cat)) {
+            return;
+        }
+        if ($this->recurse) {
+            $categoryids = question_categorylist($this->category->id);
+        } else {
+            $categoryids = array($this->category->id);
+        }
+        list($catidtest, $this->params) = $DB->get_in_or_equal($categoryids, SQL_PARAMS_NAMED, 'cat');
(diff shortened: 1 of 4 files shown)


nwp90:
Fixes #15

classes/condition/studentquiz_condition.php
```diff
@@ -37,6 +37,44 @@
...
+        }
+        list($catidtest, $this->params) = $DB->get_in_or_equal($categoryids, SQL_PARAMS_NAMED, 'cat');
+        $this->where = 'q.category ' . $catidtest;
+    }
+
+    public function where() {
+        // Gross, but rebuilds this->where with fresh catidtest...
+        $this->init();
+        $foo = parent::where();
+        error_log('studentquiz_condition returning where: ' . $foo);
```
eXpl0it3r: I assume this a left over debug call. Needs to be removed before it can be merged.

classes/condition/studentquiz_condition.php
```diff
@@ -37,6 +37,44 @@
...
+            $categoryids = array($this->category->id);
+        }
+        list($catidtest, $this->params) = $DB->get_in_or_equal($categoryids, SQL_PARAMS_NAMED, 'cat');
+        $this->where = 'q.category ' . $catidtest;
+    }
+
+    public function where() {
+        // Gross, but rebuilds this->where with fresh catidtest...
+        $this->init();
+        $foo = parent::where();
```
eXpl0it3r: Maybe use a more descriptive variable name than `$foo`.

classes/question/bank/difficulty_level_column.php
```diff
@@ -44,20 +44,60 @@ public function get_name() {
...
+        // joincondition's params.
+        $this->sqlparams = array();
+        foreach ($this->joinconditions as $joincondition) {
+            if ($joincondition->where()) {
+                $tests[] = '((' . $joincondition->where() .'))';
+                $this->sqlparams = array_merge($this->sqlparams, $joincondition->params());
+                $correcttests[] = '((' . $joincondition->where() .'))';
+                $this->sqlparams = array_merge($this->sqlparams, $joincondition->params());
+                $totaltests[] = '((' . $joincondition->where() .'))';
+                $this->sqlparams = array_merge($this->sqlparams, $joincondition->params());
```
eXpl0it3r: I don't fully understand the difference between `$tests`, `$correcttests` and `$totaltests`, can you quickly elaborate on this. If you think it's worth it, maybe even add a comment to the code.

eXpl0it3r:
[Your changes](https://github.com/frankkoch/moodle-mod_studentquiz/commit/98c1fc8b5affc444a6a0f14876e79fd85be7f6d7) have been rebased on top of my existing changes and I've removed the debugging code from your PR.

Thanks a lot again!

Issue #46, opened 2018-05-15
Anonymous setting not honoured by notification emails


mkpelletier:
It appears that the notification emails sent by student quiz include the name of commenters or question reviewers even though the activity is set as anonymous.

frankkoch:
Thank you for the report. We're aware of this but didn't see it as a bug so far. But you're right, if StudentQuiz is configured to be anonymous, there shouldn't be any loopholes about this. And if somebody wants to leave her contact data, he could do so with the comment message. So ok, it is added to our todo-list for one of the next releases.

frankkoch:
See #145

PR #47, opened 2018-05-15
Anonymize notification messages where required

mkpelletier:
Fix for #46


dionysius:
Hi @mkpelletier 

Thank you very much for your contribution. Unfortunately I can't accept this pull request in this form:
- `$anonymrank` is not available in this function signature, we have to grab this somehow from `$studentquiz->anonymrank`
- `if ($anonymrank = 0)` is always an assignment, so the `else` part will never happen
- we have also to look into the recipient and/or sender perspective according their capability (depends on message type probably) `has_capability('mod/studentquiz:unhideanonymous', $context)` if he or she is allowed to see the names despite having anonymize on
- and that the code follows coding style standards of moodle. You can find various logs in the continuous-integration/travis-ci logs (see above)

Feel free to improve the code and reopen the request. Otherwise we have it on our agenda as written by @frankkoch in https://github.com/frankkoch/moodle-mod_studentquiz/issues/46

PR #49, opened 2018-05-21
Honour anonymity settings in notifications

mkpelletier:
Fix for #46.

Sorry about that last one.  I am a beginner at moodle programming, but using an assignment operator in a condition is pretty inexcusable! :-S

Beyond the other errors you mentioned, we should also retrieve the "Anonymous" string from the lang file instead, so I did that.  Let's hope this checks out.


mkpelletier:
aw, crud.  Whitespace at end of line!  Will recommit!

PR #50, opened 2018-05-21
Honour anonymity in notifications

### LINKED ISSUE #46: Anonymous setting not honoured by notification emails

mkpelletier:
Fix #46
Add condition to mod_studentquiz_prepare_notify_data() to check for anonymity preferences in module settings.


timhunt:
On a technical note, you probably should not be adding a database query to the code here. The studentquiz data must already have been loaded somewhere, and it would be better to just pass it through to where it is needed than to query the database again.

frankkoch:
See comment from Tim

PR #53, opened 2018-06-08
Move columns display content to render

HuongNV13:
Hi Frank,
As previous conversation. I'm going to make Pull Request for phase 1 "Move columns display content to render" for that allow theme to override the content and styling for columns content

Huong Nguyen

### DIFF
--- a/classes/question/bank/anonym_creator_name_column.php
+++ b/classes/question/bank/anonym_creator_name_column.php
@@ -41,28 +41,23 @@ class anonym_creator_name_column extends \core_question\bank\creator_name_column
 
     protected $anonymousname;
 
+    protected $renderer;
+
     /**
      * Loads config of current userid and can see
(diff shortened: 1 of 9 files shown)


dionysius:
Hi Huong

We were adapting what moodle usually does in the base class. E.g. most columns print their data directly (e.g. https://github.com/moodle/moodle/blob/master/question/classes/bank/delete_action_column.php). But I also see at least a file which is using a renderer (e.g. https://github.com/moodle/moodle/blob/master/question/classes/bank/preview_action_column.php)

Optimally, we should use the renderer in a way that it doesn't include control code for decision logic (So decisions areindependent on view - they still need "if-else"s to apply decisions, but I think you know what I mean).

I don't see a point against your current proposal, I'll let @frankkoch to have a look into it.

PR #70, opened 2018-12-04
Enhancement: Ranking to include ‘Anonymised’ in block title if activity is anonymised

HuongNV13:
Hi Frank,

----------------------------
The thinking here was that we should be clear to users when the activity is or isn’t used in anonymous mode. In the future, a student might encounter a variety of anonymous and non—anonymous Student Quiz instances, and thus it should be communicated better.

Essentially IF anonymisation settings TRUE, change ‘Ranking’ title to “Ranking (anonymised)”.

This should show regardless of role, as it will remind all users of the activity as to its settings (even when moderators can see through the anonymisation setting).

So left would be what a student would see if activity is not anonymised, and how they would see it if anonymised. The only difference is that '(anonymised)' is added to the block title.
image

----------------------------
Thanks,
Huong Nguyen

### DIFF
--- a/lang/en/studentquiz.php
+++ b/lang/en/studentquiz.php
@@ -95,6 +95,7 @@ $string['no_questions_add'] = 'There are no questions in this StudentQuiz. Feel 
 // Blocks.
 $string['statistic_block_title'] = 'My Progress';
 $string['ranking_block_title'] = 'Ranking';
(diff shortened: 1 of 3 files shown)


muqiuq:
Pull request is now in branch develop4 and will be published with next release.

PR #71, opened 2018-12-04
Enhancement: Add Ranking/Stats report to Ranking/Progress boxes

HuongNV13:
Hi Frank,

--------------------------
image
**Figure 1 Admin block (staff view)**
Students do not have access to the Administration block at the OU, and therefore do not have signposted access to their reports (although they can access them if given the URL)

'More' is the usual link I've seen in Moodle, see Figure 2. 

A simple solution would be to add a second copy of the link to the bottom of the 'My progress' and 'Ranking' block. It makes more sense to be there. (Note: we are not removing the existing links from the Administration block, merely creating a secondary way to view them).

image 1
**Figure 2: mockup of new links**

--------------------------

Thanks,
Huong Nguyen


muqiuq:
Pull request is now in branch develop4 and will be published with next release.

PR #77, opened 2019-01-03
Enhancement: Question-writing and question-responding phases

HuongNV13:
Hi Frank,

----------------------------------------------------

Essentially, we need to be able to set 'start' and 'stop' dates for:
a)	Creating questions
b)	Responding to created questions.
For maximum flexibility, these should be able to work in conditions of:

- Disabled (no change to how the activity works now)
- Enabled, separate phases.
- Enabled, overlapping phases.
- Enabled, with the same time/dates (because someone will try this rather than disabling them).

This should be based on other Moodle Core tools, like Workshop, and be added to the activity set-up form, to look like Figure 1.
image
Figure 1: Student Quiz settings availability options.

**Question creation**

- If the availability criteria are not met, the ‘Create new question’ button should be greyed out.
- If before the question submission phase: a message should inform the user when they can start creating questions.
- If after the question submission phase: a message that they can no longer create questions.

**Question response**

- If the availability criteria are not met, the ‘Start Quiz’ button should be greyed out.
- If before the question response phase: a message should inform the user when they can start answering questions.
- If after the question response phase: a message that users can no longer take quizzes.

image

Figure 2

----------------------------------------------------

Thanks,
Huong Nguyen


HuongNV13:
The new code was committed to fix Page context error with Unit Test and Travis

muqiuq:
Did I understand this correctly? This new feature only disables the buttons but doesn't actually prevent "Create New Question" or "Start Quiz" to be executed?

HuongNV13:
Hi,
You mean that some Student can view or know the direct link for this function then can access the link directly?
If needed I will create extra code to throw an exception in that case

PR #87, opened 2019-01-27
removal of usage of global variables

dionysius:
From [MDL-62822](https://tracker.moodle.org/browse/MDL-62822?focusedCommentId=647722&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-647722)

> classes/question/bank/question_bank_filter.php
> Any reason for using $_POST variables directly on set_defaults()? This is not recommended at all.

> classes/question/bank/studentquiz_bank_view.php
> Another usages of $_GET and $_POST directly, pretty sure this will require some refactoring.

- removed all usage of $_ params where needed
- during the process detected, that most affected code is not necessary at all, so removed
- category is found by default category of the current context
- From the view.php: if cmid is passed via 'id', 'cmid' is getting set - this place is probably best for that need

This PR needs most likely also a good manual test of the filter! Please have a look into that too.

### DIFF
--- a/classes/condition/studentquiz_condition.php
+++ b/classes/condition/studentquiz_condition.php
@@ -81,9 +81,7 @@ class studentquiz_condition extends \core_question\bank\search\condition {
         return $this->istagfilteractive;
     }
 
-
     protected function init() {
-
         if ($adddata = $this->filterform->get_data()) {
 
             $this->tests = array();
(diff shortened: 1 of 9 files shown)


dionysius:
Update: fixed unit-test for filtering questions

- The step updateSubmission() in https://github.com/frankkoch/moodle-mod_studentquiz/pull/87/files#diff-6351ea0fef416e5c6111a6f05462ef18L127 was unneccessary, hence moodleform is capable to process the submission by itself. But it wants a valid session, which we have to fake in the unittests.

PR #107, opened 2019-04-14
Fix 106

dionysius:
I've removed all 106 related commits from master and addet them into this branch. please use feature branches.


dionysius:
cc @muqiuq

Issue #105, opened 2019-04-04
Can not display all filtered questions


miroi:
Hello,

with the fresh StudentQuiz (and Moodle 3.5) I can not display all filtered questions if they number exceed default number of questions per page.  When I click on display all question after filtering, it shows all questions in StudentQuiz, not those filtered out.

Increasing the number of questions per page does not help - after filtering it returns to the (lower) default value of 20.

Any help, please ?

dionysius:
Fixed with #112, next release coming soon containing the fix.

miroi:
ok, thanks, looking forward for the fresh release.

PR #111, opened 2019-05-13
View in one form

dionysius:
An attempt for addressing #105. The goal was one big form over the whole page. That required all actions must be buttons. Currently only the pagination is missing to be changed from links to buttons, but the class is from the moodle core...


dionysius:
Abandon this branch in favor of https://github.com/frankkoch/moodle-mod_studentquiz/pull/112. Quite a lot could be fixed, but ultimately stuck with the pagination, which used links and was not satisfyable configurable - which ultimately made all submitted POST data getting ignored and so rendering it useless. Didn't want to make the effort for an POST-friendly pagination.

PR #123, opened 2019-07-01
Automatically create Question to Studentquiz_questions table when import question from MoodleXML file

HuongNV13:
Hi Frank,

As discussed via email, this code to provide the automatically create a question in the studentquiz_questions table with default is state New when you import MoodleXML file

Thanks,


timhunt:
I just spotted this. It looks like basically the right approach.

I am just wondering one thing: in mod_studentquiz_observer, should it check that $event->contextlevel is CONTEXT_MODULE, and that $event->contextinstanceid points so a mod_studentquiz (and not something else like a mod_quiz?) You can probalby do this without a database query by using get_fast_modinfo with $event->courseid.

I hope that is helpful.

HuongNV13:
Thanks @timhunt for your comment

The fixed code was pushed to make sure that only create questions for studentquiz

Thanks,

PR #138, opened 2019-07-31
Accessibility: Filtering screenreadability

### LINKED ISSUE #137: Accessibility: Filtering screenreadability
Most of the filter options the contextual information 'Advanced element' when reading aloud by a screen reader (except Approved Questions). This information is available visually by marking form fields with a green asterisk. The screen reader is reading hidden form labels which are different from the ones that appear visually and this contextual information has not been included.

We need to include 'Advanced element' text to the hidden label of filter options for the user to know

Thanks,

HuongNV13:
resolve #137


frankkoch:
Thank you Huong

Issue #139, opened 2019-08-01
Major performance problem in the upgrade code


timhunt:
Upgrading StudentQuiz from v3.3.0 -> v4.0.1 is incredibly slow for us (20 minutes, even though we only have 10 student quizzes, and none of them contain very much data, but we have lots of quiz data in question_attempt* tables).

The problem is mod_studentquiz_get_studentquiz_progress_from_question_attempts_steps. The SELECT query there takes many minutes when I run in separately. Looknig at the 3-levels of nesting of the JOINs, I am not suprised that it is very slow.

If I can work out what this query is acutally doing, I can probably re-write it to be much faster. However, at the moment, it only makes partial sense to me. Can anyone explain?

timhunt:
That query was last changed here: https://github.com/frankkoch/moodle-mod_studentquiz/commit/14326dd22748fa7717e1f704febdf25e222afe5e#diff-528c47baf0e31a7dcd6c3042a2182b9cR180

timhunt:
Having spent some time looking at it, this query cannot possibly be right.

1) Simple error: where it joins from context to course_modules it does not check that contextlevel = CONTEXT_MODULE, so if you have a courseid that is the same as a cmid, you will get the wrong answer.

2) AND qas.state != 'todo' is wrong. Do you really mean to count thing where state is invalid?

3) The inner query 

                           SELECT MAX(qas1m.id) as attemptstepmaxid
                                FROM {question} q1m
                                JOIN {question_attempts} qa1m ON qa1m.questionid = q1m.id
                                JOIN {question_attempt_steps} qas1m ON qas1m.questionattemptid = qa1m.id
                            GROUP BY q1m.id, qas1m.userid

Runs for many minutes on our sever. It is going to compute attemptstepmaxid for essentially every questionid/userid in the system, which is many millions of rows in our database. I thinnk we should join via question_usages.contextid, and use that to limit to attempts relevant to this studentquiz.

Then, that inner join is only coupled to the rest of the query via qas1m ON qas1m.attemptstepmaxid = qas1.id, ... I guess that will ultimately limit us to one row.

timhunt:
Right, the attached commit certainly fixes the performance. The new query runs in a fraction of a second for me, whereas the old one was taking mulitple minutes.

And, I tested the numbers returned by the query for our StudentQuiz instance wiht the most attempts, and the number for that were identical. However, that is hardly a proof that the code works in all cases.

[...]

PR #141, opened 2019-08-06
StudentQuiz: fix upgrade/restore performance #139

timhunt:
With a very basic unit test.

Also removed dead code migrate.php.

Thanks to dionysius for some suggestions that have been incorporated here.


timhunt:
Last night's Travis run failed. The query being tested does not have an ORDER BY, which is good (no need to sort the results, so better for performance not to).

However, this caused the unit test to fail randomly. I have now changed to unit test to be reliable.

The remaining Travis failure above is spurious (the VM failed to initialise properly). If re-run, it should go green.

timhunt:
Last night's Travis run failed. The query being tested does not have an ORDER BY, which is good (no need to sort the results, so better for performance not to).

However, this caused the unit test to fail randomly. I have now changed to unit test to be reliable.

The remaining Travis failure above is spurious (the VM failed to initialise properly). If re-run, it should go green. (It passed for me: https://travis-ci.org/timhunt/moodle-mod_studentquiz/builds/568783737).

dionysius:
I tried to fix the travis behat issue, but it's kinda difficult to prepare the affected table. lets go with `| true` for now

timhunt:
Thanks!

The Behat problem is being worked on in https://tracker.moodle.org/browse/MDL-66378, so hopefully we can remove the | true bit again soon.

PR #143, opened 2019-08-13
When Rating/Comment enforced, visual cue

HuongNV13:
You can now enforce Rating and Commenting (or not).

However, you only find out if it is enforced if you try saving/moving on without having rated or commented.

So the easiest thing is to be consistent with other required fields in Moodle and apply the image of the red asterisk (e.g. the "Title" field in the SQ instance set-up page).

Expected:
image

Thanks,


frankkoch:
Thanks Huong

PR #148, opened 2019-08-19
Change State when answering/previewing question need some improvements

### LINKED ISSUE #147: Change State when answering/previewing question need some improvements
The 'Submit' button needs to be greyed out until the user change the State select box

The 'Submit' button needs to be greyed out during the State change processing. Being able to 'Submit' the same State repeatedly should not be allowed

Loading icon will be shown during the State change process to let the user know their request is being processed

StudentQuiz State Change

Thanks,

HuongNV13:
resolve #147


frankkoch:
Thank you Huong

PR #149, opened 2019-08-20
StudentQuiz with Moodle 3.6

### LINKED ISSUE #146: Automatically create Question to Studentquiz_questions table when import question from MoodleXML file do not work with Moodle 3.6
Just found this.

The reason is **\core\event\question_created** event only available from Moodle 3.7.

Finding a solution

HuongNV13:
- Automatically create Question to Studentquiz_questions table when import question from MoodleXML file do not work
 - Loading icon do not work

resolve #146


HuongNV13:
StudentQuiz 3 6

Testing with Moodle v3.6

frankkoch:
Thank you Huong

PR #158, opened 2019-08-22
Difficulty bar not work with PosgreSQL Database

### LINKED ISSUE #156: Difficulty bar not updated
Under Moodle prior to v3.7:
1) create a question
2) run as quiz and provide wrong answer for this question. As expected, you get a difficulty bar with 'my difficulty' = 100%
3) run the quiz again. This time provide a correct answer for the same question. Since now there is 1 correct and 1 incorrect answer, 'my difficulty' should be updated to 50%, but it ramains on 100%.

This error doesn't appear under Moodle v3.7

HuongNV13:
resolve #156


frankkoch:
Thank you Huong!

PR #162, opened 2019-09-06
Email header should display anonymous student and manager if needed

nashtechdev01:
Email header should display anonymous student and manager if needed.

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -289,6 +289,7 @@ function mod_studentquiz_prepare_notify_data($question, $recepient, $actor, $cou
     $data->actorusername = $recepient->username;
 
     // Set to anonymous student and manager if needed.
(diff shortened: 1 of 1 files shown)


nashtechdev01:
Hi @frankkoch,
Please help me to review and merge to your repo. This commit has been tested on local.

Thank you,

dionysius:
Hi, I've not forgotten this PR. I wanted to take the time and get rid of some discrepancies in that part of code. Especially
- Get rid of any `firstname` and `lastname` usages, only name - since moodle offers `fullname()` for users so we as well should only use fullnames when there should be written something anonymous in that
- Unify translation templates by getting rid of `username` duplicate and only using `actorname` and `recipientname`

PR #168, opened 2019-11-18
Create a default category if it is missing

rschrenk:
We had the issue that there was no category in the question bank for this module. Therefore the code needs a case for this.

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -69,6 +69,17 @@ function mod_studentquiz_load_studentquiz($cmid, $contextid) {
             if ($studentquiz->category = question_get_default_category($contextid)) {
                 $studentquiz->categoryid = $studentquiz->category->id;
                 return $studentquiz;
+            } else {
+		// If we come here we have no question category for this module.
+                // Consequently we create a default category.
+                require_once($CFG->dirroot . '/question/tests/generator/lib.php');
+                $cqg = new core_question_generator();
+                $cdata = array(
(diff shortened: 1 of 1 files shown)


dionysius:
Hi @rschrenk! Thank you for your contribution!

I can't accept this PR because it is using test code from a foreign module (and also watch for the moodle coding styleguide which made the check to fail).

And I'd like to know why and how it happened that your studentquiz has no category.

rschrenk:
Hi @dionysius 

What do you mean with "test code from a foreign module"? I only use the API of moodle core.
I too have no idea why there was no question category by default.

The check fails because the code is indented by tabs instead of spaces. I pasted the code from Atom into the textarea on github. I think this caused the issue, but this can be handled easily ;-)

Kind regards

locallib.php
```diff
@@ -69,6 +69,17 @@ function mod_studentquiz_load_studentquiz($cmid, $contextid) {
             if ($studentquiz->category = question_get_default_category($contextid)) {
                 $studentquiz->categoryid = $studentquiz->category->id;
                 return $studentquiz;
+            } else {
+		// If we come here we have no question category for this module.
+                // Consequently we create a default category.
+                require_once($CFG->dirroot . '/question/tests/generator/lib.php');
```
dionysius: That part is test code, so we can't use that maintainable.

PR #170, opened 2019-11-20
Need to stop same-state resubmits

HuongNV13:
Hi Frank,
I have a small change as below:

 - We need to stop same-state resubmits. After the state was changed successfully, the user only able to submit the new state if the new state is different with last changed state.
 - Also, after the state was changed successfully, the parent window (StudentQuiz page) will be reloaded to update the latest state of that question

Thanks,
Huong Nguyen


frankkoch:
Thank you Huong

PR #175, opened 2019-12-17
Move 'Publishing' setting from Ranking to General settings section

HuongNV13:
The 'Publish new questions' settings is the second setting under the 'Rankings' subsection of the activity settings.

As a general setting, it should be in 'General' settings, above Ranking


dionysius:
This is already in the recently merged PR #180 with it's improvement, so this got obsolete.

PR #179, opened 2019-12-31
StudentQuiz\Comments: make it into a proper 'discussion area'

lamtranb:
We'll want to make additional improvements to the Comments area. 

As discussed, the additional improvements should be:
- Text area for inputting comment/response is an editor with Bold, Italic, superscript, subscript, link/unlink, bullet list, numbered list, and 'HTML view'.
- Expanded/collapsed feature. Default is collapsed view displays 5 latest comments.
- Show reply, delete button and count replies.
  + Reply button only show for top-level comment, not response comment. 
  + For students, delete button only show for 10 minutes. Deleted comment is hided from student and shown to admin/tutor in grey.
-  If ‘Commenting’ is enforced, it will accept either a top-level comment OR response comment.
- In anynomous mode, username will be hided from students (not hided from their own comments).


dionysius:
Hi @lamtranb, thank you very much for the contribution! I've discussed the outline with @frankkoch and this has been overall approved, but I'd like to ask you for the following:

- if a comment is deleted and the user is not allowed to view these, still show an entry that a comment has been deleted (e.g. https://qph.fs.quoracdn.net/main-qimg-76f128e36a56de4797473217e66f10f0.webp)), otherwise response history might be questionable for a future visitor.
- Make the allowed deletion period configurable (you can still leave 10min as default). We shouldn't add _magic constants_, especially since there's a probability that it affects the real life teaching style we can't know. Probably for now an entry after `Enforce commenting` is sufficient. If there are more settings related to commenting coming I would then move them to their own section.

Would you like to to that? Otherwise I can also pick that up but then it might take a while until I can tackle onto this.

HuongNV13:
Hi @dionysius,
Lam is my colleague and he is working with me on many tickets related to StudentQuiz in OU.

I have one question related to the deletion period configurable setting. Is this a per-instance setting or global setting?

Thanks,

dionysius:
I think per instance would be sufficient, but this is also a good example for both. Many settings in studentquiz use both global and instance settings, where the global one sets the default for the instances and so still allows to be overridden by the manager/teacher if necessary. For example `forcecommenting`: global https://github.com/frankkoch/moodle-mod_studentquiz/blob/11630b9436334d2b3377150ab77c79f0fd267ac6/settings.php#L122 (use admin_setting_configtext() with INT probably, e.g. see `studentquiz/addquestion`), and in the instance with default from global https://github.com/frankkoch/moodle-mod_studentquiz/blob/master/mod_form.php#L197

dionysius:
I see you still push changes to this feature. Can you comment here when you think I can consider a review and merge of this feature? Thank you again for the idea, I like it a lot - and for the contribution!

HuongNV13:
We are doing the test for this feature. We will let you know when the code is ready.

Thanks,

dionysius:
just to inform you, dont worry about behat tests failing on moodle 38: see https://github.com/blackboard-open-source/moodle-plugin-ci/issues/109

HuongNV13:
I'm going to create another PR to include 3 others enhancement related to this feature.
@dionysius , could you please help me to close this?

PR #184, opened 2020-01-14
WIP: UI, Modal state change not working

dionysius:
1) fixed: Submit button selector not working in boost
2) open: select element doesn't get the change event 90% of the time

I've found the reason, why the change event never gets triggered. By looking into the jquery event list for that element, we can see, that sometimes the change function is assigned (and then works), and mostly not (and obviously doesn't work).

I assume we have here some javascript timing/loading issues. Maybe because it *is* a modal? @HuongNV13 Do you maybe have an idea?

Screenshot from 2020-01-14 19 07 48
vs
Screenshot from 2020-01-14 19 07 13


HuongNV13:
I'm facing the same problem after upgrading to v4.0.5 for OU.
The reason is the new Boost theme generated as below (button type=submit, not input type=submit). That's why the selector does not work. In other themes, it generates input type=submit.

image

Fixed code committed and tested.

Thanks,

dionysius:
Hi Huong, I already found and did that in this PR, so https://github.com/frankkoch/moodle-mod_studentquiz/commit/0fb65f6da779474506fb33278da553a007d3ea5c and https://github.com/HuongNV13/moodle-mod_studentquiz/commit/f6e5ed6dcc81548b8ba589a1ff7855d359f02108 do the same.

I'm curious why I have issues with the select change event. How often do you have a result in `jQuery._data($('#menustatetype')[0], 'events')`. Because locally I have that only ~10% working

HuongNV13:
In my case, it works normally all the time

dionysius:
Well, I merge this and we'll see if we get some feedback. Maybe I'm just unlucky on my local installation

iamwretched:
We're going to update to fix this, but as far as I can see it worked OK in FireFox but not in Chrome in the January version of StudentQuiz. Just for your information.

PR #189, opened 2020-01-20
Improve auto set-up of question capabilities when a StudentQuiz is created

### LINKED ISSUE #182: Improve auto set-up of question capabilities
@timhunt was so kind and commented a review with suggestions
https://github.com/frankkoch/moodle-mod_studentquiz/commit/8823472fe7e642d17b4b91376288d04d585e5528#commitcomment-36792684

HuongNV13:
Thanks @timhunt for the comment,

I already fixed the code and created Unit Test for this case

This will resolve #182 

Thanks,

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -1048,21 +1048,66 @@ function mod_studentquiz_get_roles() {
 function mod_studentquiz_ensure_question_capabilities($context) {
     global $CFG;
 
-    $neededcapabilities = array(
-        'moodle/question:add',
-        'moodle/question:usemine',
-        'moodle/question:viewmine',
-        'moodle/question:editmine'
-    );
+    $neededcapabilities = [
+            'mod/studentquiz:view' => [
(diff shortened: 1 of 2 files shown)


timhunt:
In the latest commit d674c79b5780ce29ece6b7827039b26814e73666, the code in mod_studentquiz_ensure_question_capabilities looks spot on to me.

I don't understand the unit test.
* It is not at all obvious what the unit test is testing.
* Also, how can a single unit test be sufficient? There are lots of possible code paths through mod_studentquiz_ensure_question_capabilities, and to test even just the few most important cases (e.g. Student and Teacher) surely requires more than one test?

Anyway, the OU would quite like to get these changes, so I don't really want my concerns about the unit tests to delay this further. As I say, I think the actual mod_studentquiz_ensure_question_capabilities code looks right. The unit test can be improved later.

PR #194, opened 2020-02-07
StudentQuiz: Behat tests failing due to how data is passed to JS

lamtranb:
Hi all,

Sometimes behat will fail if behat test site has long domain name. 

> Too much data passed as arguments to js_call_amd("mod_studentquiz/comment_area", "generate"). Generally there are better ways to pass lots of data from PHP to JavaScript, for example via Ajax, data attributes, ... . This warning is triggered if the argument string becomes longer than 1024 characters.

I have fixed it. Please help me review it.

Thanks.


lamtranb:
Hi all,

Behat works fine with my travis build
[https://travis-ci.org/lamtranb/moodle-mod_studentquiz/builds/648277084](https://travis-ci.org/lamtranb/moodle-mod_studentquiz/builds/648277084)

But it does not seem to work here
[https://travis-ci.org/frankkoch/moodle-mod_studentquiz/builds/648216470](https://travis-ci.org/frankkoch/moodle-mod_studentquiz/builds/648216470)
> Selenium server is not running, you need to start it to run tests that involve Javascript. More info in https://docs.moodle.org/dev/Running_acceptance_test

Please help me review.
Many thanks

PR #204, opened 2020-03-02
there is no need to limit the results by question category if we limit the subresult already by question id

### LINKED ISSUE #195: Tags not visible when questions are in a subcategory
Hello,
with the most recent StudentQuiz v4.1.1 2020013000 the tags in questions created by students are no longer visible in the StudentQuiz main question page if the question happens to be in a subcategory of the StudentQuiz.

However, when the question is moved from a subcategory to the main StudentQuiz category, tags are visible again. Maybe this is the purpose ?

dionysius:
fixes #195 

@timhunt @HuongNV13 Anyone likes to have a quick review here?

### DIFF
--- a/classes/condition/studentquiz_condition.php
+++ b/classes/condition/studentquiz_condition.php
@@ -23,6 +23,8 @@
 namespace mod_studentquiz\condition;
 defined('MOODLE_INTERNAL') || die();
 
+require_once($CFG->dirroot . '/mod/studentquiz/classes/local/db.php');
+use mod_studentquiz\local\db;
 
 /**
  * This class controls from which category questions are listed.
@@ -78,12 +80,6 @@ class studentquiz_condition extends \core_question\bank\search\condition {
         return $this->isfilteractive;
     }
 
-    protected $istagfilteractive = false;
-
-    public function is_tag_filter_active() {
-        return $this->istagfilteractive;
-    }
-
     protected function init() {
         if ($adddata = $this->filterform->get_data()) {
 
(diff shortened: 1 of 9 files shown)


timhunt:
I think I see two issues with the patch:

1) JOIN {question} q ON ti.tagid = q.id looks like the wrong JOIN condition.

2) This is only a worry that I have not tested, but with this sort of subquery, you need to worry about performance. Some stupid databases will not optimise the query correctly unless you give them some help. E.g. MySQL might evaluate the inner query we are building here for ever questionid in the database, which on a large database could take forever.

However, the query before your patch, with the subquery in the ON, is not best for performance either. In fact, I worry about the whole approach used here. I think you should not try to use get_extra_joins like this at all. Instead do it all in get_required_fields.

(Note, you MUST use place holders to put values into SQL, or you risk SQL injection attacks.)

public function get_required_fields() {
    return array('(
        SELECT COUNT(1)
          FROM {tag} t
          JOIN {tag_instance} ti ON t.id = ti.tagid
         WHERE ti.itemtype = 'question'
           AND ti.itemid = q.id
    ) AS tags', '(
        SELECT COUNT(1)
          FROM {tag} t
          JOIN {tag_instance} ti ON t.id = ti.tagid
         WHERE ti.itemtype = 'question'
           AND ti.itemid = q.id
           AND t.name LIKE :searchtag
    ) AS tagssearchtag');
}

The looks like horrible duplicate code, but I am pretty sure query optimisers will handle this much, much better.

dionysius:
Bad news first: I was almost able to get rid of the left join subquery. But I hit a dead end when I wanted to try to filter the result by the topmost query alias which was constructed as `GROUP_CONCAT(t.name) AS tagarray`. [MySQL doesn't allow aliases to be used in where](https://stackoverflow.com/questions/942571/using-column-alias-in-where-clause-of-mysql-query-produces-an-error). I am unsure about the HAVING one, I can't predict any performance impact on that.

Back to the topic, I was able to get rid of a looot of stuff and I also expect a performance boost by getting rid of separate queries. Basically:
- the tag list is now directly read via DB, so the database can make optimisations
- this way, all the parts which would read the tag list after separately could be removed
- several functions in the *column classes were now unused and so removed as well
- I had to create an additional database driver since `GROUP_CONCAT` is mysql specific.
- still have to test on postgres

Would like to have a rereview :)

timhunt:
Note that Moodle core recently (Moodle 3.8 and above - https://tracker.moodle.org/browse/MDL-66553) added display of question tags in the question bank. This commit: https://github.com/timhunt/moodle/commit/61cc1e64947fc8ef4db79ea99f1b490fad3f2684

So, as soon as we move to Moodle 3.8 being the minimum supported version, that will make this easier. But, I guess we are not in that situation yet.

timhunt:
If you can't refer to the alias tagarray, just put in the expression GROUP_CONCAT(t.name). Databases will spot that it is the same expression, and evaluate it efficiently (I am pretty sure).

PR #206, opened 2020-03-02
Exclude hidden questions in question resultset if the user is not allowed to view them, instead of just not showing them.

### LINKED ISSUE #201: Additional pages not visible if approval of questions is required
We encountered an issue with a StudentQuiz with question approval required. New, unapproved questions filled the first page of the question list. Now students were unable to see any questions, even though there were multiple pages of approved and visible questions. Even the navigation bar for question pages was gone. The problem can be hotfixed by having at least one approved question on the first page, but this behaviour does not seem to be intended. Is it possible to pre-sort the list by visible and invisible items and put the visible ones in front?

dionysius:
Fixes #201 

Fixes also issues with navigation because of that (empty question page).

@HuongNV13 can you review the changes please? You provided that code part initially

Review Notes:
- Changed the "LEFT JOIN" to a "JOIN", so it now limits the Resultset
  - That's why most other parts could be omitted so I removed them as well

### DIFF
--- a/classes/question/bank/sq_hidden_column.php
+++ b/classes/question/bank/sq_hidden_column.php
@@ -63,14 +63,12 @@ class sq_hidden_column extends action_column_base {
      * @param string $rowclasses CSS class names that should be applied to this row of output.
      */
     protected function display_content($question, $rowclasses) {
-        if (has_capability('mod/studentquiz:previewothers', $this->qbank->get_most_specific_context())) {
-            if ($question->sq_hidden) {
-                $url = new \moodle_url($this->qbank->base_url(), ['unhide' => $question->id, 'sesskey' => sesskey()]);
-                $this->print_icon('t/restore', get_string('restore'), $url);
-            } else {
-                $url = new \moodle_url($this->qbank->base_url(), ['hide' => $question->id, 'sesskey' => sesskey()]);
-                $this->print_icon('t/show', get_string('hide'), $url);
-            }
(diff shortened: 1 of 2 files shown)


timhunt:
Comments similar to the one I made of the other pull request apply.

I think that in get_extra_joins you should do something like

```
public function get_extra_joins() {
         return ['sqq' => ' JOIN {studentquiz_question} sqq.questionid = q.id'];
}
```

And then

```
public function get_required_fields() {
        return ['sqq.hidden AS sq_hidden'];
}
```

Oh, I am wrong. That is not quite the right logic. You need to test in the SQL, but only if they don't have the capability. Well, that should be done with

```
public function get_extra_joins() {
         return ['sqq' => ' JOIN {studentquiz_question} sqq.questionid = q.id AND sqq.hidden = 0'];
}
```

Except you need one method with the appropriate if statement to choose  between these to implementations.

(Sorry, this comment is  not very clear, but I hope this is enough of a hint.)

HuongNV13:
For the get_extra_joins() function, I think it's fine to me.

But for the display_content() function at https://github.com/frankkoch/moodle-mod_studentquiz/blob/926ebf9d526fdbce6cf815882588e9f6a989ff96/classes/question/bank/sq_hidden_column.php#L65

The old code with the capabilites checking, if you don't have the capability, the Show/Hide button will not visible. With the new one, the buttons will be visible all the time.

PR #211, opened 2020-03-26
Incorrect comment deletion period helptext

HuongNV13:
Currently, the help text always said that the default value is 10 minutes, even the default setting was changed in the Admin setting


dionysius:
We should do that the moodle way with `$a`: e.g. https://github.com/frankkoch/moodle-mod_studentquiz/blob/master/lang/en/studentquiz.php#L70

PR #239, opened 2020-05-28
Move Atto Toolbar config to admin setting to suppport Atto WordCount feature

HuongNV13:
Currently, the atto toolbar of comment editor was hard coded to allow only a few plugins.
It should be moved to Admin setting, that help us to easily allow which plugins will be shown in the toolbar

Thanks,


timhunt:
In my personal opinion, StudentQuiz should just use the standard Atto toolbar config.

HuongNV13:
I agree with you, but last time Chris want to limit the plugins. So I decided to create the setting

Issue #249, opened 2020-06-15
Student can 'Edit' and 'Delete' question of others after remove capability 'previewother'


HuongNV13:
1. Login as Admin/Manager
2. Go to StudentQuiz.
3. Create question (True/False).
4. Create student account.
5. Allow the "previewother" capability for that student account.
6. Log in as student.
7. Go to StudentQuiz above and verify that student can preview/edit/delete the question.
8. Login back to admin.
9. Disallow the "previewother" capability for that student account.
6. Log in as student again and go to StudentQuiz above
7. Verify that student cannot preview/edit/delete the question.

Actual: Student still can see the edit and delete button

dionysius:
You can see the actual permissions by role if you go to the StudentQuiz -> gear icon -> check permissions. Keep in mind that those permissions are only in that context if you added those permissions in the context!

After adding and removing the previewothers permission we are at:

![image](https://user-images.githubusercontent.com/1341084/84713207-caadbf80-af6a-11ea-809d-7b43b450e42e.png)
![image](https://user-images.githubusercontent.com/1341084/84713226-d7321800-af6a-11ea-8046-1b5e82e3fe48.png)

As we can see, the critical permissions in questions got automatically added. I also think those should also not be added at all! This is probably due to https://github.com/frankkoch/moodle-mod_studentquiz/blob/98252597b7b42e4c90bd0c187a9c65f33ea6273f/locallib.php#L1006.

Independent of what capabilities are set, the more important question is also: how to get the "inherited" capabilities automatically removed? It looks like removing a studentquiz capability does not remove the "inherited" capability

dionysius:
If a student can preview others, he should definitely not see the option to change the state **and not be able to actually change the state!**

![image](https://user-images.githubusercontent.com/1341084/84714283-87088500-af6d-11ea-90a7-1a20ca30fa23.png)

dionysius:
I started a branch permission and I found an additional thing to discuss about (See current diff: https://github.com/frankkoch/moodle-mod_studentquiz/compare/permissions)

previewothers capability should IMO not allow for changing states. It should only allow previewing other questions. That leads to the next question: Changing a state of a question is something that non-editing teachers would do, while "manage" is the perfect capability for editing teachers, how should we assign them a capability to do that?

dionysius:
I propose the following:

- preview.php needs a require_capability previewothers
- introduce a new capability changestate
[...]

PR #251, opened 2020-06-16
Context specific permissions are not revoked if the studentquiz capability is removed

dionysius:
Implemented according discussion and suggestion in https://github.com/frankkoch/moodle-mod_studentquiz/issues/249


dionysius:
This PR does not improve the way how often this ensure function is called, I would like to implement a way to listening on events. So far I see, that we would need to listen on:
- added/removed capability event (so existing enrolled roles in the course get the update)
  - but as the function currently states, we need to be careful that we can filter the events, otherwise we create an event loop. Which eventually solves itself - in theory - but I don't know Moodle well enough. We better find a way to filter that.
  - The event a user is enrolled to the course, since for get_roles_used_in_context only roles are returned, which users represent in the course

But I suggest that the event listening part is added by a new PR based on this PR. @timhunt would you mind having a review on this? Thank you in advance

timhunt:
What you have done looks generally good to me. However, it is the end of a long, very hot day, here. So I am not sure how well my brain is working!

I agree with doing the event listening part in a future PR.

The following are the things I noticed.

When defining new capabilities, you can specify 'clonepermissionsfrom'. This means that if people have custom roles set up, you can set a new capability like an exisiting one, which might be even better defaults than the archetypes. E.g. it might be good to clone organize from manage.

Putting $contextcapabilities in db/access.php feels wrong to me. I agree that it is good to move this array to somewhere central. I guess if it was me, I would be gradually moving code out of locallib.php and into auto-loaded classes. So, I might make a class like mod/studentquiz/classes/permission_helper.php, and put all the code related to this, including this array, in there.

Calling a DB function like $DB->get_field in a loop is bad. All pages should be O(1) DB queries. However, deleting questions is hardly the most perforance-critical thing. Might be better to leave this for now. (But please make a new issue so we don't forget. We can probably get Huong to fix that.)

We really should not be using $_REQUEST, but I see that is an existing issue, so now is not the time. Acutally, the way we handle it with mod_studentquiz_helper_get_ids_by_raw_submit is about as good as it can be.

Will be interesting to see what the automated tests say.

I hope those comments are useful.

timhunt:
Also, please may I add a really huge thank-you to you for working on this.

dionysius:
Thank you for your quick response! I hope you can cool down a bit 😄 The suggestions are definitely helpful!

That one big chunk of change just looks like much, but I actually haven't changed what the code does (except adding the right capability check) - I wanted to reduce duplicate code in that function so it is easier to simplify the code in the future.

dionysius:
Forgot to ask: Do you know if I can use `clonepermissionsfrom` without setting archetypes or do both need to exist?

timhunt:
> Forgot to ask: Do you know if I can use `clonepermissionsfrom` without setting `archetypes` or do both need to exist?

You need to set both.

Archetypes are used when someone installs StudenQuiz for the first time.

clonepermissionsfrom will be used only during upgrades, if the other permission you refer to allready existed before the upgrade.

dionysius:
Changes since last review: https://github.com/frankkoch/moodle-mod_studentquiz/compare/3ad5610...permissions

PR #260, opened 2020-08-17
Permissions event

### LINKED ISSUE #249: Student can 'Edit' and 'Delete' question of others after remove capability 'previewother'

dionysius:
Based on #251 so better use https://github.com/frankkoch/moodle-mod_studentquiz/compare/permissions...permissions_event for comparison.

Moves the actual call to code to apply context specific capabilities (overrides) to event observers, which reduces the amount of calls drastically (before was on every page load of every user, now only on event changes that matter). Fixes #249 and I consider I am complete with the changes. @timhunt do you mind having a look at this as well? Thank you in advance!


dionysius:
Hmm... Moodle 35 and 37 don't have the events we need:

| Event | 35 | 37 | 38 | 39 |
| --- | --- | --- | --- | --- |
| \core\event\capability_assigned | :heavy_multiplication_x: | :heavy_multiplication_x: | :heavy_check_mark: | :heavy_check_mark: |
| \core\event\capability_unassigned | :heavy_multiplication_x: | :heavy_multiplication_x: | :heavy_check_mark: | :heavy_check_mark: |
| \core\event\user_enrolment_created | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| \core\event\user_enrolment_updated | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| \core\event\user_enrolment_deleted | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| \core\event\course_module_created | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |

That means we need to conditionally do the old way... argh

timhunt:
Note, the error from travis looks real (missing global $CFG). Also, Moodle HQ have taken over maintenance of the Travis integration, and it woudl be good to upgrade some time: https://moodlehq.github.io/moodle-plugin-ci/UPGRADE-3.0.html

Now, to acutally review the code.

timhunt:
I think that, overall, this now looks like it should work. Good work.

Minor pedantic comments:

1. There are rules for PHP namespaces in Moodle https://docs.moodle.org/dev/Coding_style#Parts_of_a_namespace. Would be better to follow them.
2. Similarly function names like ensurerelation should be ensure_relation
3. Since $studentquizrelation is a property of class contextoverride, my choice would be to not pass it in as an argument to ensurerelation.
4. I don't see invoke_method being used in the unit tests. Should it be deleted?

dionysius:
I agree with all those points. One question thought for point 3: Where do I place local "config"? I've designed context_override to be reusable, even for external modules. But if I want to make a "separation of concerns" (which is here meant the separation of data to execution code) I can't figure out where moodle expects me to place the relation "config" which is some sort of "internal config". Is there a "localconfig.php" file allowed to be used?

I mean `classes/*` is the better version of `locallib.php`. `db/*` is a reserved dir by the guideline itself. That's why I added it there for now since I didn't know a better place.

PR #261, opened 2020-08-26
Fix debug code related to Digest setting

HuongNV13:
- Do not execute this event if we change from No digest to Daily/Weekly. This will fix: Undefined variable: timetosend
 - Use the correct format for strtotime function. This will fix: Notice: A non well formed numeric value encountered

Thanks,


timhunt:
https://github.com/frankkoch/moodle-mod_studentquiz/pull/261/commits/6b71549054fbdf6ccc41d032591319fa98f8db89?w=1 is a clearer link to review. Change makes sense to me, but I don't know this area of the code very well.

PR #262, opened 2020-08-26
Sometime event cannot bind to State change combobox

HuongNV13:
Only bind the combo box event once the page loaded


timhunt:
https://github.com/frankkoch/moodle-mod_studentquiz/pull/262/commits/1f8375811c2b7da52f8d3477e68b8ea17c5f22e3?w=1 makes the change easier to see. Looks right to me.

PR #277, opened 2020-10-13
fix #275, allow a user to see his own question, even when hidden

### LINKED ISSUE #275: students do not see their own questions after saving
We operate an instance of Moodle 3.8 which is hosted centrally by the German state, so we have no access to the database. A great tool so far. Since the update from 3.7 to 3.8 when creating a new StudentQuiz in version 4.3.2 (old variants from 3.7 no longer exist), students do not see their own questions after saving, regardless of the filter setting. Students can only edit it again after the teacher has confirmed the question and then changed it to changed.

The steps to understand:

1 Click as student “Create new question” and create any question,
2 Filling in the question, title test question,
3 Save,
4 List shows no entry for test question, regardless of the filter,
5 teacher sees new question test question and confirms and then changes to changed,
6 only then a student can edit his question.

However, the behavior as described in https://tube.switch.ch/videos/8194aa08 from 8:20 min is desirable. What can I do?
krueuw

dionysius:
The question bank column classes really need an overhaul to **correctly escape** input data!


dionysius:
This PR should simplify the use of the "current" user and studentquiz id in the question bank column queries. First it improves maintainability by not needing to define those all the time, and secondly in case additional sanitation has to be made, it is in a well defined place.

Moodle has the option to use `:variablename` in queries, but unfortunately you cannot use that multiple times. This way, you have a joined "table" `current` providing those fields as static select elements and you can always use those as conditions for further joins. I expect the performance also to be as the database can earlier filter the joined tables according to the join conditions. But that statement is just from expectation, there's a chance that I'm totally wrong and its worse and thus needs to be tested as well.

dionysius:
```
TEST_PERF: create 1 instances
TEST_PERF: initialization complete
TEST_PERF: displaying question bank took 0.21114706993103 s
TEST_PERF: create 10 instances
TEST_PERF: initialization complete
TEST_PERF: displaying question bank took 0.52533197402954 s
TEST_PERF: create 100 instances
TEST_PERF: initialization complete
TEST_PERF: displaying question bank took 0.23308610916138 s
```

But this approach doesn't like pgsql. Putting PR aside for a bit

timhunt:
I don't see why this would not work in Postgres. Perhaps you need to add `ON 1 = 1`?

In studentquiz_bank_view.php, you should be able to do `JOIN (SELECT ? as userid, ? as studentquizid) current` and put those values in the $params array.

dionysius:
> I don't see why this would not work in Postgres. Perhaps you need to add ON 1 = 1?

I can give that a try.

> In studentquiz_bank_view.php, you should be able to do JOIN (SELECT ? as userid, ? as studentquizid) current and put those values in the $params array.

Yes definitely, now we can do that.

PR #286, opened 2020-11-02
Support adaptive question types

### LINKED ISSUE #213: Last attempt in adaptive mode
Hello,

I use the StudentQuiz module, which allows students to use a mechanism for synthesizing knowledge through questions in a competitive and social environment. I love this module.

However, when trying to use the CodeRunner module in which questions are run in adaptive mode (as they provide immediate feedback), the last answer to the question is always marked as always wrong.

I attach an image that indicates this behavior.

Thank you for your support!

dionysius:
This PR gives StudentQuiz support for adaptive question types like CodeRunner (fixes #213).

StudentQuiz prefers when the question type supports immediate feedback and provides such option to the question engine when constructing the quiz (which later forwards that to the question type).

While many core Moodle question types can do both modes, they behaved with immediate feedback, because StudentQuiz preferred it to do so.

StudentQuiz now supports question types forcefully doing adaptive. I tried to keep the "adaptiveness" of the questions by "finishing" these types of questions only when navgating further - allowing a student to improve the answer before the attempt is recorded.

The effective score doesn't matter, just the resulting state after "finishing" the question.

This support needed some tweaks to detect that a question has been at least answered in the current attempt, because rating, commenting and overall navigating were dependent on that state.

I suggest to also rewrite the rating save code by removing save.php and making a service target for adding a rating.


timhunt:
I just had a quick look at the changes here. I can't say that I have done a full review, but they seem like they are are probably right.

dionysius:
Thank you for your review. I've not observed any double attempts of the same questions manually and is also proven by the tests. The most flaky part might be the detection of `isanswered` if a plugin qtype for whatever reason already adds another attempt step than the first one automatically added by the question engine.

I really like to do at least a test with some question being forcefully adaptive, but we don't have much ready in the unit tests regarding a running quiz and it's quite difficult to do in behat. My best bet is currently behat where I have a small version of run_quiz.feature where I set question(s) with moodle core qtypes forcefully to adaptive using a custom behat expression.

My biggest question to you: Is the defined "adaptiveness" sufficient to you? Currently it finishes the question only when the student clicks on Abort, Next or Finish. The attempt is only recorded when the question got finished. If someone just closes the tab, this attempt is lost.

timhunt:
Ah! you should definiely NOT look at the sequence count. E.g. in CodeRunner the Precheck button will add a step.

You should just look at $laststate. If $laststate is question_state::$invalid or question_state::$todo, then it is not done, otherwise it is. ... I think. ... No, that is not right. Adaptive only goes to 'Complete' state when you have got the question right.

OK, how about this: 
```
$hasanswered = false;
switch($questionusage->get_question_state($slot)) {
    case question_state::$gradedpartial:
    case question_state::$gradedright:
    case question_state::$gradedwrong:
    case question_state::$complete:
        $hasanswered = true;
        break;
    case question_state::$todo:
    default:
        $hasanswered = false;
}
if (!$hasanswered && method_exists($questionusage->get_question_attempt($slot)->get_behaviour(), 'get_graded_step')) {
    $hasanswered = null !== $questionusage->get_question_attempt($slot)->get_behaviour()->get_graded_step();
}
```

That should work for adaptive and adaptive_adapted_for_coderunner

dionysius:
I really appreciate your help! It feels so good to get my thoughts out to someone before I merge stuff. Back to the topic:

Unfortunately, I've seen it first hand on my local installation, if someone answers CodeRunner wrongly, the **state of the question is still todo**! :disappointed: Otherwise I would've already implemented a form of your suggestion.

timhunt:
Sorry, I gave no explanation of why my suggested code might be sensible.

The point is, knowing what the steps in an attempt really mean is something that only the behaviour can do.

Now, at a basic level, that meaning is communicated thorough the question_attempt state, but that is optimised for what the quiz navigation needed when this was first implemented. Still, for almost all question behaviours, it works for us.

The ony behaviour is is not really working for are the adaptive-derived ones, and those behaviours all seems to have a get_graded_step method, which will tell us what we want to know: has a response been submitted for grading successfully. So, if other tests fail, we look to see if that method exists, and if it does, we call it.


Incidentally, in your code that hard-codes the test for essay and descriptions. Again, it would be theoretically more correct to look at the behaviour, not the question type. That is, you should do something like

```php
$behaviour = $questionusage->get_question_attempt($slot)->get_behaviour();
if ($behaviour instanceof qbehaviour_informationitem || $behaviour instanceof qbehaviour_manualgraded) {
   // ....
}
```

This would work for third-party questoin types that work like essay or description. E.g. our https://moodle.org/plugins/qtype_recordrtc.

dionysius:
Thank you for the explanation! The code of StudentQuiz is often from a big while ago where we or the previous devs didn't find out all the quirks. So it's always very valuable to have you here :+1: 

I'll update the detection of the first answering and the exclusion of "non-functional" questiontypes

dionysius:
- `$behaviour instanceof qbehaviour_adaptive` guarantees that this class must have a method `get_graded_step()`
- allowed qtypes are filtered by `is_real_question_type()` and `is_manual_graded()`

PR #299, opened 2020-12-12
prepare for PHPUnit 9

ewallah:
Moodle 3.10 is preparing to use PHPUnit 9.

The proposed changes work only in Moodle 3.8+


ewallah:
PHP fatal errors make it impossible to run PHPunit testing:

PHP Fatal error:  Declaration of mod_studentquiz_bank_performance_test::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void
PHP Fatal error:  Declaration of mod_studentquiz_permissions_testcase::tearDown() must be compatible with PHPUnit\Framework\TestCase::tearDown(): void

timhunt:
Thanks for looking at this.

If you do this, you need to follow Moodle coding style. Space after the colon in ): void {.

Apart from that, these changes look safe (the don't stop the tests running with recent earlier Moodle versions which have not upgraded the PHPunit used.) I have done similar things in many of my plugins.

timhunt:
Would be great to fix this. @ewallah, will you have time to fix your patch to follow Moodle coding style?

timhunt:
Sadly, this is causing PHPunit fails on the 3.5 branch: https://travis-ci.org/github/frankkoch/moodle-mod_studentquiz/jobs/762096793. E.g.

```
1) mod_studentquiz_cron_testcase::test_send_no_digest_notification_task

Error: Call to undefined method mod_studentquiz_cron_testcase::assertStringContainsString()
/home/travis/build/moodle/mod/studentquiz/tests/cron_test.php:141
/home/travis/build/moodle/lib/phpunit/classes/advanced_testcase.php:80
```

Do we still need to support 3.5? (I think I ahve stopped for most of my pluigins. 3.9 is the most recent LTS core Moodle version.)

timhunt:
Thanks @dionysius. That is fail. However: 

* If a particular Moodle site is not upgrading their Moodle code, will they be upgradeing the StudentQuiz plugin installed in it?
* And, will they ever run the unit tests?
* Most importantly, unit tests are used by people doing development, and those people tend to be using recent Moodle versions. 

For example, OU moved to Moodle 3.10 in our developments yesterday, and now our test runs are failing because of this. (Annoyingly, this does not just make the one test fail, it causes the PHP process runnign PHPunit to crash out completely.)

So, we would vote to merge this sooner.

dionysius:
I think I wasn't clear what I meant to say. Conclusion was: That doesn't mean we have to stick to this. We can't argue about what organisations do and do not, we would only assume. Fact is, moodle 3.5 gets security updates until May 11th. It's about the decision we make to give both sides the best and the change to do so is clear.

I propose to deprecate 3.5 (and 3.7) and add 3.10 with StudentQuiz 4.5.x and onwards, and keep 4.4.x as security updates for Moodle 3.5 if neccessary.

Right now 3.10 has not yet been added to our Readme and Travis pipeline, because if I remember correctly, neither `mdk` nor `moodle-plugin-ci` had support for it as it was still in beta at that time. But now it's long overdue to to so.

Proposed steps as PR to master:
[] Remove support and code specific to moodle <= 3.7
[] Officially support Moodle 3.10 by updating Readme, travis, and etc. where necessary
[] Merge this PR
[] Release v4.5.0

timhunt:
Moodle plugin-ci has supported 3.10 for a while.

That is a sensible plan.

Note that Travis-CI is getting less and less good. Most people are moving to github actions instead, which seem to be more reliable, and better integrated with github. See https://moodlehq.github.io/moodle-plugin-ci/#github-actions. So, would be worth doing that.

I have had a busy week this week. There is more chance I might have time to help next week.

dionysius:
I can start with that

dionysius:
I've started a branch `sq45` (not pushed yet), where I merged in this PR and wanted to start with switching to Github Actions.

I just found out, that github actions doesnt allow to exclude matrix combinations, (only include)[https://github.community/t/improve-matrix-exclusion-feature/16340/6]. Will play around with it

timhunt:
I've always felt that if I am getting someone else's CPU resources for free, it is good to be respectful, therefore I normally just use include to run a sample of different Moodle versions with different databases and PHP versions, like this: https://github.com/moodleou/moodle-qtype_pmatch/blob/main/.github/workflows/ci.yml

(But, other people like to do it differently.)

dionysius:
Closing this. I've opened PR #316 which includes this PR and further discussion and implementation should happen there. This PR is accepted.

Issue #203, opened 2020-02-21
Cache related (?) time delay for students permissions


cdipe:
We have a very odd problem that we haven't got in any other moodle component. Also, we cannot replicate the problem in our QA environment, which is on a more simple server system than the PROD environment. The PROD is on a REDIS cluster with four server nodes.
Using moodle 3.6.6 and StudentQuiz 4.1.1

1. The teacher adds a StudentQuiz
2. Directly after a student enters the StudentQuiz, but gets the message "Sorry, but you do not currently have permissions to do that (access question edit tab questions)."

Now, first we added the permission **moodle/question:editall** for students in the course, and the students could use the StudentQuiz. I then discovered that if we instantly removed the permission again, it still worked for the students. We just forced some other switch when manipulating the permission.
Here's a short video showing the procedure: https://play.lnu.se/media/t/0_4qx9x61t

3. We noted that without manipulating anything, just adding a new StudentQuiz and wait, after some time has passed (10-15 minutes, sometimes less) the student could use the StudentQuiz without any problem. That seemed cache related to us.
4. So, we tested purging all caches instantly after having added a StudentQuiz, and directly after having done so the student could use the StudentQuiz. Obviously the permissions I changed earlier somehow had the same effect.

But purging the cache didn't solve the issue; when adding another StudentQuiz we still have to wait a while for that "something" to happen before it works for the students, unless we purge the cache again or manipulate the permissions.

So, do you know what this could all be about? Is there anything specific involving the StudentQuiz and the cache?

frankkoch:
Thank you Clipe for the report and the video. The error might be caused by some JavaScript that has not been initialised before being called. I checked the permission settings and they work fine. So I think the permission have nothing to do with the observed behaviour. Its rather a matter of time.

cdipe:
Thank you @frankkoch, I'll ask our technicians to check the javascripts out. Exactly, it's a matter of time but I haven't been able to find any consequent pattern to when it "solves itself". Sometimes after 10 minutes, next time it works directly, next time i takes hours (unless any manual workaround action is taken). I'll get back if we find a solution to our problem.

dionysius:
[...]

PR #309, opened 2021-02-18
Improve role sync performance. Fixes #306

### LINKED ISSUE #306: mod_studentquiz incredibly slowing down our site
Hello,

we are using certain roles on the system and certain course category contexts. We experienced that processing role assignments and unassignments was getting slower and slower by the time. Recently we ran into timeout errors and made a deeper analysis what was going on.

In this example, one user is assigned 1 role and unassigned one role at the same time.

We found out, that the observer of mod_studentquiz is called every time, when a role changes. As you can see, it consumes extremely much of the time for mapping the roles to override capabilities.

mod_studentquiz

I am not sure what the purpose is to check for role assignments above the course context. IMHO mod_studentquiz only needs to consider course enrolments and course roles.

Could you please consider to (1) introduce a new role that grants the required capabilities and is assigned to all users automatically on the level of the course module based on course enrolments, and (2) use ad hoc tasks so that these assignments (that are not time critical) are performed by the cronjob?

Kind regards

timhunt:
Before this comment, as soon as any role was changed, the permissions for
all student-quizzes were changed.

After this, when the roles change, then we note that time in a cache.

Then, when a user tries to access a StudentQuiz, if the roles have
changed since we last synced permissions for this StudentQuiz,
then we re-sync them (and note the fact in the cache).

Because of the way the cache is used, if the caches are purged, then
we will re-sync the permissions for each StudentQuiz the next time a
user tries to access it.

### DIFF
--- a/db/events.php
+++ b/db/events.php
@@ -46,13 +46,5 @@ $observers = [
                 'eventname' => '\core\event\capability_unassigned',
                 'callback' => 'mod_studentquiz_observer::capability_unassigned',
         ],
-        [
-                'eventname' => '\core\event\role_assigned',
-                'callback' => 'mod_studentquiz_observer::role_assigned',
-        ],
-        [
-                'eventname' => '\core\event\role_unassigned',
-                'callback' => 'mod_studentquiz_observer::role_unassigned',
-        ],
         // I assume course_module_deleted removes anyway all capability overrides, so not added here.
 ];
--- a/db/events.php
+++ b/db/events.php
@@ -54,9 +54,5 @@ $observers = [
                 'eventname' => '\core\event\role_unassigned',
                 'callback' => 'mod_studentquiz_observer::role_unassigned',
         ],
-        [
-                'eventname' => '\core\event\course_module_created',
-                'callback' => 'mod_studentquiz_observer::course_module_created',
-        ],
         // I assume course_module_deleted removes anyway all capability overrides, so not added here.
 ];
--- a/attempt.php
+++ b/attempt.php
@@ -51,6 +51,10 @@ $slot = required_param('slot', PARAM_INT);
 $attempt = $DB->get_record('studentquiz_attempt', array('id' => $attemptid));
 
 $context = context_module::instance($cm->id);
+
+// Check to see if any roles setup has been changed since we last synced the capabilities.
+\mod_studentquiz\access\context_override::ensure_permissions_are_right($context);
+
 $studentquiz = mod_studentquiz_load_studentquiz($cmid, $context->id);
 
 global $USER;
--- a/classes/access/context_override.php
+++ b/classes/access/context_override.php
@@ -77,6 +77,47 @@ class context_override {
         ],
     ];
 
+    const ROLES_CHANGED_TIME_CACHE_KEY = 'roleschanged';
+
+    private static function cache_key_for_cm(int $cmid) {
+        return 'cm' . $cmid . 'synced';
+    }
+
+    /**
+     * This method gets called by the observer class whenever roles change somewhere.
+     */
+    public static function roles_setup_has_changed() {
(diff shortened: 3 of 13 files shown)


timhunt:
Hopefully the Amend I jsut pushed will fix the Travis failure.

HuongNV13:
This looks very good, Tim.
I agree that since we call **ensure_permissions_are_right**() function every page where a user interacts with a StudentQuiz, we can stop to listen the **role_assigned** and **role_unassigned** events.


Just some minor points related to wording:
https://github.com/frankkoch/moodle-mod_studentquiz/commit/4806d601f13670b6f0b45376135e62a120283254#diff-7ce0a1c1ce9cad4274da94d5d3b04542513c86e5d57e08f7a3ea4528926280b3R110
Should it be "and even on one server"?

And (If we still keep these even listeners)
https://github.com/frankkoch/moodle-mod_studentquiz/commit/4806d601f13670b6f0b45376135e62a120283254#diff-38a8807f90f0ee92d3bbe9ecaf36eb165e825541fb514b12ddca072620fc0f41R149
Should it be "not need to listen for role_assigned events"?

https://github.com/frankkoch/moodle-mod_studentquiz/commit/4806d601f13670b6f0b45376135e62a120283254#diff-38a8807f90f0ee92d3bbe9ecaf36eb165e825541fb514b12ddca072620fc0f41R161
Should it be "not need to listen for role_unassigned events"?

timhunt:
Thanks Huong. Typos fixed.

We cannot yet stop listending to role_assigned/role_unassigned, because of this line: https://github.com/timhunt/moodle-mod_studentquiz/blob/roles_perf/classes/access/context_override.php#L139

At the moment, permssions are only fixed in roles where someone is assigned them. We would need to change that to fix all roles when this function is called, then there would be no need to worry about roles being (un-)assigned.

timhunt:
I added a second commit to the branch, which alwasy fixes all roles, and therefore does not need to listed to the role_assigned events. I am pretty sure that will be a big overall performance win (many fewer calls to ensure_relation).

dionysius:
We can't remove the role_[un]assigned events. [Un]assigning roles to an existing enrolment only calls those events. This way a new role can be introduced or removed to all studentquizzes in this course.

Screenshot from 2021-02-19 20 35 06

dionysius:
Otherwise looks good to me. I was hoping not to need to introduce a call on every page. How does this change affect #203, since we now actively introduce a specific activity-level MUC cache key?

db/events.php
```diff
@@ -46,17 +46,5 @@
...
                 'callback' => 'mod_studentquiz_observer::capability_unassigned',
         ],
-        [
-                'eventname' => '\core\event\role_assigned',
-                'callback' => 'mod_studentquiz_observer::role_assigned',
-        ],
-        [
-                'eventname' => '\core\event\role_unassigned',
-                'callback' => 'mod_studentquiz_observer::role_unassigned',
-        ],
```
dionysius, line 56: we can't get rid of those, see my comment https://github.com/frankkoch/moodle-mod_studentquiz/pull/309#issuecomment-782301910

timhunt:
We can get rid of those.

What ensure_relation does is, for each role, it sets the moodle/question:* capabilities based on the mod/studentquiz:* capabilties.

If that is right for each role separately, then you don't need to worry if a user has a role assigned or unassigned.

In the past we did need to worry, because ensure_relation() only did its thing for roles where at least one user had the role in this StudentQuiz. Therefore, the releationship had to be re-checked if a role assignemnt was added or removed. Now, I have changed ensure_relation() to process all roles whenever it is called. Therefore, you don't need to worry about which roles are assigned.

------

We don't need to worry about calling ensure_permissions_are_right lots. For almost every call, all it will do is to fetch two ints from a cache, and compare them. This is very fast. It will only do work after a role has been edited.

PR #316, opened 2021-03-16
WIP: Preparation for StudentQuiz 4.5.0

dionysius:
- [x] Remove support and code specific to moodle <= 3.7
- [x] Officially support Moodle 3.10 by updating Readme, CI, etc. where necessary
- [x] Officially support Moodle 3.11 by updating Readme, CI, etc. where necessary
- [x] Merged PR https://github.com/frankkoch/moodle-mod_studentquiz/pull/299 for PHPUnit 9 compatibility
- [x] Switch to Github Actions

Help appreciated. Pick an item and set a PR against this branch `sq45` (or `master` is also fine, I'll manually merge it into here then)

PR #317, opened 2021-04-29
Added support for SQL Server 2017 and newer

mrkskwsnck:
Since SQL Server 2017 and newer has the `STRING_AGG` function, it can be treated like PostgreSQL. Also, SQL Server is not able to `GROUP BY` column aliases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Supported and tested with:
 
 - Moodle 3.5, 3.7, 3.8, 3.9
 - PHP 7.1, 7.2, 7.3, 7.4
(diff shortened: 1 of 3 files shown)


dionysius:
Thank you! That was smaller than I thought

mrkskwsnck:
I′m not sure what to do here. Do I need to recommit my changes with PSR2 compliance?

dionysius:
No, the error does not look related to your PR. We're switching to Github Actions in #316 and will include this PR for the 4.5.0 release

dionysius:
included in PR #316

PR #318, opened 2021-05-18
Comment sorting only affects the latest 5 comments

JBThong:


dionysius:
The pipeline error PSR2.Methods.MethodDeclaration.StaticBeforeVisibility can be ignored, I fixed this already in another PR

JBThong:
Hi @frankkoch,

Currently, the sorting function only sorts the five latest comments, even if the question has more than five comments.
The sorting function should do the sort base on the total comments of the question, not only for the last five comments.
I have fixed this issue in this pull request. Could you help me review it?

Thanks.

timhunt:
I think that the fix does not go far enough. With this change, I think there is no longer having the first if ($haslimit) branch in query_comments. I think the sort order should always be the same, however many comments are being fetched.

Apart from that, I think the patch looks correct.

JBThong:
Hi @timhunt, 

Thanks for your review, I have removed the condition ($haslimit) branch in query_comments. I think it looks better. Could you please help me take a look at it?

Best regards,

JBThong:
Hi Tim,

Thanks for your review, I have removed the condition ($haslimit) branch in query_comments. I think it looks better. Could you please help me take a look at it?

Best regards,

PR #328, opened 2021-08-12
StudentQuiz: Sticky question.

JBThong:


JBThong:
Hi Team, 

In this feature, I have created a sticky icon to pin/unpin the question on the top table. This feature has been approved by Frank.

Thanks,

timhunt:
For reviewing this, it would be helpful if you could attach some screen-grabs.

timhunt:
Thanks for working on this Thong. Most of your change looks absolutely spot-on. All the database/permissions checks, etc.

My main concern is about the usability. I think using the star/unstar icons for the action to pin or unpin is confusing. Pinning by teacher is very different from a user starring a question for themselves. I think that the User interface for this should be as close as possible to the pinned post feature in the standard Moodle forum.

And, related to this, I am not sure it makes sense to have two separate columns. Why not combing the action of pinning and unpinning with the display of the pin?

We probably need to discuss this. I send a meeting request for Monday.

JBThong:
Hi Tim, 

I have fixed your feedback. I have attached the screen-grab. Could you please help me look back at it again?

Thanks.

https://user-images.githubusercontent.com/32395146/132463573-aab997c8-c869-4c53-a6d6-14b852e7786a.mp4

timhunt:
After the last review, we had a meeting, and agreed the following changes:

1. Keep current UI approach that we have 2 columns star/undstar and pin/un-pinning
2. We are planning a future change (#330) to replace the row of icons with an edit menu including the pin/un-pinning feature and existing actions ( delete, zoom, setting,..), like core question bank. At this point, we can have words, which helps explain the slighly cryptic star/unstar icons for pin/unpin, so we won't worry about those icons for now.
3. Thong will add small space/ margin between Pin icon and Name question

These changes have now been done, so merging.

PR #329, opened 2021-09-13
Studentquiz: Behat failure

JBThong:


timhunt:
This is a nasty fix (delays in tests always are) but I guess we live with it for now. (I think the issue that needs to be fixed is in the core comments API.

Merged.

PR #330, opened 2021-09-16
Studentquiz: Add menu action

JBThong:
Hi @timhunt,

I have improved the menu action in this commit. All actions of the question such as "edit, delete, show/hide, preview" will be moved to the menu action.
I have attached 2 images below:

Thanks
image
image


dionysius:
I like this - there got more and more buttons in the past, so a menu fits well.

Personal optinion/question: Shouldn't we put this column to the end of the table? Because I've rarely/never seen actions to be in between the table and since we don't try to be similiar to the question bank table anymore.

timhunt:
The position of the menu is an intersting question. In the standard Moodle question bank, the menu is next to the question name, becuase the actions relate to the question.

Checking a few other systems I use (Azure DevOps bug tracker, SharePoint file storage) they also have the action menu next to the name of the thing. So, my opinion is that the Edit menu is in the best place currnetly, putting it right beyond all the meta-data would be less usable.

timhunt:
I am a bit puzzled.

I asked a question "Why is any custom CSS required?" I was expecting you to write a reply here.

I was not expection you to develop a lot more code (which involved a huge amount of copying and pasting, so is clearly worse than a bit of CSS.) Please can we acutally discuss this - perhaps after stand-up.

nguyenphuctien:
Sorry for annoying, Tim.

As I said, Thong is adding the 'Pin question' into the action menu. We adding the css to fix an accessibility issue (when the question is hidden, the accessibility of the action menu is not good). However, your feedback is totally right. We shouldn't do it that way. We also ask Chris about the color of the Edit link via email, and do it in the right way.

I think that Thong is using the git action to check his work (while it's uncompleted) and it cause a bit annoying.
Hi @JBThong, could you please another branch to do it next time because whenever you push a commit, everyone will get the notification?

JBThong:
Dear Tim,

I apologize for my mistake. I see that my actions impacted you. I should reply and discuss with you before back to the development.
In the future, I will take care of this, so that this situation won’t repeat itself.

Best regards.

JBThong:
Thanks for your information @nguyenphuctien,

It's clearly detailed. I have got your advice on this issue too.  I will follow it.

nguyenphuctien:
Hi @timhunt,
Thong has rebased and remove the custom css. What he's actual did is remove the dimmed_text class in the row (tr) and put it into the columns (td) instead: https://github.com/JBThong/moodle-mod_studentquiz/commit/141c1a47093bf6ea5508a712452b0628af1c5fe2

Thong has just squash all and the new code is passed local test and git action checks.

Could you please help us to review it again?

nguyenphuctien:
anonym_creator_name_column extends \core_question\bank\creator_name_column directly (old code),
the same with question_name_column extends \core_question\bank\question_name_column.

If we extend them directly from studentquiz_colum_base, we need to copy some code from  \core_question\bank\creator_name_column and \core_question\bank\question_name_column.

PR #340, opened 2021-10-21
Improve web accessibility

taboubi:
Here some screen shots:
The contrast of numbers's color with the background is low (Checked with Wave)

EVOSTDM-2520-c

Focus on buttons:
EVOSTDM-2520-b

Table overflow
EVOSTDM-2520-d

### DIFF
--- a/renderer.php
+++ b/renderer.php
@@ -119,28 +119,22 @@ class mod_studentquiz_renderer extends plugin_renderer_base {
         $bc->content = html_writer::div($this->render_progress_bar($info1), '', array('style' => 'width:inherit'))
              . html_writer::div(
                 get_string('statistic_block_progress_last_attempt_correct', 'studentquiz')
-                .html_writer::span('<b class="stat last-attempt-correct">' .$userstats->last_attempt_correct .'</b>', '',
-                    array('style' => 'float: right;color:#5cb85c;')))
+                .html_writer::span($userstats->last_attempt_correct, '', ['class' => "stat badge last-attempt-correct"]))
             . html_writer::div(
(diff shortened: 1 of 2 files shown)


renderer.php
```diff
@@ -262,10 +256,10 @@ public function render_progress_bar($info, $texttotal=null, $bicolor=false) {
 
         // Stylings.
         $rgbstroke = 'rgb(200, 200, 200)';
-        $rgbyellow = 'rgb(255, 193, 7)';
-        $rgbgreen = 'rgb(40, 167, 69)';
+        $rgbyellow = '#ffc107';
+        $rgbgreen = '#28a745';
```
timhunt, line 260: Please follow the style of the existing code, and user rgb(...).

styles.css
```diff
@@ -792,3 +796,35 @@
     vertical-align: middle;
     margin: 0 9px 0 0;
 }
+
+.path-mod-studentquiz .createnewquestion {
+    margin: 0.5rem 0;
+}
```
timhunt, line 802: In Moodle now, the preferred approach is to use Boostrap classes, rather than custom CSS, if possible, which itis here. https://getbootstrap.com/docs/4.0/utilities/spacing/

viewlib.php
```diff
@@ -295,8 +295,7 @@ public function get_context() {
      * @return string
      */
     public function get_title() {
-        return get_string('modulename', 'studentquiz') .
-                ': '.  $this->get_coursemodule()->name;
+        return $this->course->shortname . ': ' . format_string($this->get_coursemodule()->name);
```
timhunt, line 298: What is the purpose of this change? I don't see it in your description.

PR #350, opened 2021-10-26
Limit the roles to show in the parameter of the activity

taboubi:
* Add "Roles to show" parameter in the admin plugin setting
	* In the list of roles to exclude from ranking in the parameter activity, only show the roles selected in "Roles to show"
rolestoshow
activityparam


taboubi:
Hi @timhunt , I made some changes.

timhunt:
Thanks for continuing to work on this.I think we are very, very close, but there are still some things to fix. (Since a rebase is needed, hopefully it is not too much extra work to fix the other two trivial points.)

1. Other changes have just been merged, so I think you will have to rebase again. Sorry.
2. It has not been necesasry to use =& for objects in PHP since PHP 4 went out of support. There is one in the form class (which came from existing code) but let's fix it now.
3. Please consistently use [] for arrays in new code. (You got most of them, but there is at least one array() in reportlib.php, and one in the unit test.) Please only update ones in the code you have already changed in your patch.

Thanks.

taboubi:
Done

timhunt:
Great. Thank you very much.

PR #351, opened 2021-11-01
Remove update_active_tab_api and update accesibility for question preview tabs

nguyenphuctien:
We have things worth to fix to make the code better from #341:

1. Moodle has a standard (if slighly old) way for JavaScript code to set user preferences. See M.util.set_user_preference in lib/javascript-static.js. Creating a whole webservice in our plugin for this seems excessive.
=> I have removed the redundant webservice and changed to using M.util.set_user_preference.

2. Moodle has a standard way to render table (core_renderer::tabtree()). For UI consistency, it would be better to use that, rather than to maintain our own template.
=> I have checked, and the core_renderer::tabtree() do not support bootstrap 4.0 tabs (https://getbootstrap.com/docs/4.0/components/navs/#tabs). It just renders a tabs with links only, so we have to create a template in studentquiz. I hope that Moodle core will create a core component for nav-tabs soon. I've added arial-selected to support accessibility for it.


timhunt:
Thanks :-)

PR #352, opened 2021-11-08
Add anotate_ids for statehistory backup stebslib

nguyenphuctien:
We are missing annotation for userid and questionid. If we backup an studentquiz and restore, then we preview the question and view the state history, sometime it will throw an error
Screen Shot 2021-11-08 at 4 36 10 PM


timhunt:
Thanks. Merged.

PR #354, opened 2021-11-10
Studentquiz: Sort by time in State history

JBThong:
Hi Tim, 

I have added the order clause for timecreated field.

Thanks
image

### DIFF
--- a/classes/utils.php
+++ b/classes/utils.php
@@ -727,7 +727,7 @@ style5 = html';
     public static function get_state_history_data($question): array {
         global $DB;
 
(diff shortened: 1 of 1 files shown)


timhunt:
For a computer, 1 second is a long time and several things can happen. Please can we make it

'timecreated, id'

so the order is always absolutely predictable. (No needs for ASC, that is the default.)

Apart from that, this change is fine.

JBThong:
Yes, Tim. I have updated.

Could you please help me review it again?
Thanks

timhunt:
Thanks. Merged.

PR #355, opened 2021-11-11
Disable buttons when there are no selected questions

nguyenphuctien:
In this commit, I have fix some issues:
1. Disable all action buttons when no questions are selected.
2. In case people cheat by typing directly in the url, I have added the code to redirect after remove the action from the parameters if:
- We enter the change sate confirmation screen or delete confirmation screen without any selected question.
- We move to category without any selected question
3. In the behat tests, I have added @javascript to make sure the js code to click on the "check all" button will be called before we check the disable attribute of buttons.

Hi @timhunt,

could you please help me to review it?

Thanks.


timhunt:
Thanks.

PR #357, opened 2021-11-17
Studentquiz: Fix code review in ticket add state history table

JBThong:
Hi @timhunt, 
As your review in the PR:  https://github.com/studentquiz/moodle-mod_studentquiz/pull/346
I have improved:

1. delete_data_for_user in the privacy. I have changed the user id to guest id for the question with state new because the owner of the question is the quest. With another state such approved, disapproved, changed, hide, show. I have changed the user id to admin id.
2. In the fix_all_missing_question_state_history_after_restore function, I have used the get_recordset instead of get_records to improve performance.
3. In the upgrade.php. I have changed get_recordset to get_records and added a progress_bar.
progress_bar

Thanks

### DIFF
--- a/classes/privacy/provider.php
+++ b/classes/privacy/provider.php
@@ -545,6 +545,7 @@ class provider implements
         $user = $contextlist->get_user();
         $userid = $user->id;
         $guestuserid = guest_user()->id;
+        $adminid = get_admin()->id;
 
         list($contextsql, $contextparam) = $DB->get_in_or_equal($contextlist->get_contextids(), SQL_PARAMS_NAMED);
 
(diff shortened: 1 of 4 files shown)


timhunt:
Thanks Thong. That is a really big step in the right direction.

I think there is just one thing that needs to be improved more. In the classes/utils.php code, you do one SQL query, and then loop over the results, and do a $DB->count_records (which really ought to be a record_exists). Anyway, I think all this can be combined into one query:

```sql
SELECT sqq.questionid, sqq.state, q.createdby, q.timecreated
                      FROM {studentquiz} sq
                      JOIN {context} con ON con.instanceid = sq.coursemodule
                      JOIN {question_categories} qc ON qc.contextid = con.id
                      JOIN {question} q ON q.category = qc.id
                      JOIN {studentquiz_question} sqq ON sqq.questionid = q.id
                     WHERE sq.coursemodule = :coursemodule
                           AND qc.id = :categoryid
                           AND NOT EXISTS (SELECT 1 FROM {studentquiz_state_history} WHERE questionid = q.id)
```

(Only change in that query is adding the last line.)

Can you make that additional change?

JBThong:
Yes, Tim. I have updated the SQL query. 

Could you please help me to review it again?
Thanks

timhunt:
Thanks Thong.

PR #358, opened 2021-11-19
Fix inconsistency division in columns

nguyenphuctien:
We use a pipe ( | ) divider in "My attempts" to differentiate between the two data, so we should possibly do the same for "Comments".
Screen Shot 2021-11-19 at 2 01 45 PM


timhunt:
Thanks.

PR #359, opened 2021-11-24
Add indices to speed up privacy requests

t-schroeder:
This speeds up privacy requests that fetch data about a user. Those are currently taking forever in our production system due to studentquiz.

### DIFF
--- a/db/install.xml
+++ b/db/install.xml
@@ -124,6 +124,9 @@
         <KEY NAME="commentid" TYPE="foreign" FIELDS="commentid" REFTABLE="studentquiz_comment" REFFIELDS="id"/>
         <KEY NAME="userid" TYPE="foreign" FIELDS="userid" REFTABLE="user" REFFIELDS="id"/>
       </KEYS>
+      <INDEXES>
+        <INDEX NAME="userid" UNIQUE="false" FIELDS="userid"/>
+      </INDEXES>
     </TABLE>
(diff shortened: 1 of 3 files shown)


dionysius:
looks good to me

timhunt:
@t-schroeder, thanks for looking at StudentQuiz performance.

However, your proposed changes should have no effect. Althought Moodle does not create acutal forign key constraints in the database, it does automatically created indexes where a column is used in a foreign key declaration.

I just checked in the database for my Moodle site, and those indexes already exist.

Did you do any permormance measurements before and after adding the indexes? (If so, it would have been good to share them in your pull request.)

If you go to Admin -> Development -> XMLDB editor in your Moodle site, and run [Check indexes], what does that say about the StudentQuiz tables?

Anyway, I am going to close this.

t-schroeder:
@timhunt thanks for the explanation. I hadn't tested this. I just saw in install.xml for the table studentquiz_comment that there was both a foreign key as well as an index. So I thought I'd have to do the same for these other columns to improve their performance.

https://github.com/studentquiz/moodle-mod_studentquiz/blob/979c914c7a8e6ca7eb2b25f2c866c88c331beb9b/db/install.xml#L88-L112

timhunt:
Well spotted. That is a bug that should be fixed. I will make a new issue for that.

timhunt:
Issue #362

PR #361, opened 2021-11-25
Studentquiz: Display current state when changing states

JBThong:
Hi @timhunt 

In this commit, I have displayed the current states of questions when changing states.
I have attached 2 images below:
1. In the dialog form change states:
popchangestates
2. In the preview page:
preview change states

Thanks


timhunt:
This looks good to me. @dionysius, any comments before I merge it?

dionysius:
Looks good to me if we want to keep effort low. The previous way was not clever and efficient and would've been a chance to do so. But not at all required. Looks good to me

dionysius:
Unrelated thought: what does question in use even mean for the user? From: https://github.com/studentquiz/moodle-mod_studentquiz/blob/master/lang/en/studentquiz.php#L304
> (* Questions marked by an asterisk are already in use in some quizzes.)

Why does it matter for the user? We can probably just remove the fancy display of in-use asterisks, since it is a technical situation related to question deletion and independent to states.

timhunt:
@dionysius I'm sorry, I don't see any "review comments" other than your comments here. Am I missing something?

I don't think the "question in use" is related to this change, so I think we shoud not address it here. We are going to have a lot of work in relation to the Moodle 4.0 release, so I think we should deal with that then.

dionysius:
Ah, i need to click the submit review button.

> I don't think the "question in use" is related to this change, so I think we shoud not address it here.

Totally fine for me, just thoughts ahead

JBThong:
Thanks, @dionysius, and @timhunt

I will update the points which you mentioned:

1. $questionnames .= $DB->get_field(…) to prevent multi-call DB queries in this ticket. 
2. About render_current_state_questions function, I will change the render loop like that through questions and render out those after each other, also removing the <br /> tag as your suggestion.
3. "question in use" things I will create a new issue to fix it later?

Many thanks,
Thong

dionysius:
Wait a bit until @timhunt comments, my comments were just suggestions which could be unneccesary to put effort in.

timhunt:
Thong and I were discsussing this offline :-)

JBThong:
Hi @dionysius 

I have updated the points you mentioned above in this [commit](https://github.com/JBThong/moodle-mod_studentquiz/commit/bd007a89c25c510c7a759d5248a81ded92aa9e09). I have squashed commit too.
Could you please help me review it again?

Thanks,

dionysius:
yeah, looks good to me. Thank you!

timhunt:
Thanks Thong. I will merge this now.

PR #363, opened 2021-11-30
Changes after code base switch.

lucaboesch:
You should probably get rid of Travis CI and old repository references in README.md.
Also, indicate compatibility with Moodle >= 3.8 in the badge, as in the text further down.


timhunt:
Thanks Luca.

PR #364, opened 2021-12-02
sped up queries for data requests

t-schroeder:
The long OR conditions don't work well with indices. Therefore the
original statements got split up into several statements that each just
look at one condition.

### DIFF
--- a/classes/privacy/provider.php
+++ b/classes/privacy/provider.php
@@ -148,58 +148,51 @@ class provider implements
 
         // Get activity context if user created/modified the question or their data exist in these table
         // base on user ID field: rate, comment, progress, attempt.
-        $sql = "SELECT DISTINCT ctx.id
-                  FROM {context} ctx
-                  JOIN {studentquiz} sq ON sq.coursemodule = ctx.instanceid
-                       AND contextlevel = :contextmodule
-                  JOIN {question_categories} ca ON ca.contextid = ctx.id
-             LEFT JOIN {question} q ON q.category = ca.id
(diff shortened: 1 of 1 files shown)


t-schroeder:
As opposed to #359 this fix has been tested, even on our production system, and speeds up data requests from 1.5h to 4min.

The only side effect of this change is that the data comes in a slightly different order. See [data-original.txt](https://github.com/studentquiz/moodle-mod_studentquiz/files/7640719/data-original.txt) vs [data-after-change.txt](https://github.com/studentquiz/moodle-mod_studentquiz/files/7640720/data-after-change.txt). But I think the order isn't prescribed anyway.

timhunt:
Thanks for working on this.

Several points:

1. Please ensure your changes don't break CI.
2. If your changes, you have unnecessarily changed the whitespace in the SQL. This makes it difficult to see what has really changed. Please don't do that. (Or, of you want to improve coding style, please do that changes in a separate commit - but please ensure they are genuine improvements, rathern than just changing perfectly well laid out code to match your personal preferences.)
3. If we are no longer querying everything at once, do we need to keep the monster join? Could it instead use fewer simpler queries?
4. One feature of the old, single query, is that it avoided duplicates in the export. Does the new code ensure that? Is there a unit test to verify it?
5. Or, is the new code only ensuring this by pulling all the data into memory, which might lead to out-of-memory errors on large sites?

timhunt:
Closing this. Please re-open, or create a new pull request when there is something ready to review.

PR #365, opened 2021-12-08
Activity-specific setting to enable private discussions

nguyenphuctien:
I have created this pull request to add activity-specific setting to enable private discussions:

On the Comment and rating settings for StudentQuiz, Adding new setting call "Enable private commenting" below "Enforce commenting" with
- Default value: get "Enable private commenting" setting from site setting
- Help text: The private discussion thread, accessible when Previewing a question until a question is Approved, allows discussion between the question author and the activity administrator(s). If the StudentQuiz mode is set to 'Automatically publish new questions', then the question author will not necessarily ever see this specific thread. However, it remains available to the activity administrator(s) regardless, who might use it to explain why they have hidden or deleted a question. The private commenting default is set in the server-level plugin settings.
- This setting will support enable private commenting that follow standard setting rule from Moodle (sequence priority activity level, course level)
- Support Moodle standard: Backup, restore, accessibility.
Screen Shot 2021-12-08 at 10 18 55 AM

Default value Enable private commenting setting for StudentQuiz:
- Existing SQ will be upgraded that doesn't have new "Enable private comment" setting : get "Enable private comment" setting value from site setting (currently, default  site setting = off)
- Backup/ restore SQ that doesn't have new "Enable private comment": get "Enable private comment" setting value from site setting (currently, default  site setting = off)


timhunt:
Thank you for the clear description fo what you have done above. That makes this much easier to review.

The code looks good to me.

@dionysius, do you have any comments before I merge this?

PR #370, opened 2022-01-06
Function return types

marcusgreen:
Tim: I have seen your message to say you have your own plans, but since I have written this I'll make it available and you can ignore it or I will decline/remove/delete it or whatever. It is purely to add some return types that would be necessary for 4.0 and should work with the current version.


timhunt:
I'm going to close this. It will get dealt with as we fully fix 4.0-compatibitity.

Also, you might like to review the recommended coding style for this.

PR #374, opened 2022-01-19
Localize weekly and daily in mail digest notification fixes #373.

### LINKED ISSUE #373: "Daily" and/or "Weekly" appear untranslated in e-mail notifications
"Daily" and/or "Weekly" string is passed to the mailing mechanism without being localized resulting in phrases such as 

Dies ist Ihre Daily Ausgabe der Benachrichtigungen für die eHealth StudentQuiz StudentQuiz-Aktivität, hier verfügbar:https://moodle.moodle.moodle/mod/studentquiz/view.php?cmid=1425290

The code that causes this is here, probably: https://github.com/studentquiz/moodle-mod_studentquiz/blob/master/classes/task/send_digest_notification_task.php#L77

Best,
Luca

lucaboesch:
This might fix the "Daily" and/or "Weekly" appear untranslated in e-mail notifications issue.
I think that it is also more appropriate to have daily and weekly written in lower case since they are in midst of a sentence.

But it might be that you want to resolve this in a different way. ~~Note that in non-english languages as well as with adopted lang strings the URL link sticks directly to the colon since the string 'This is your {$a->digesttype} digest of notifications for the <b>{$a->modulename}</b> StudentQuiz activity, available here: '; can't be translated in AMOS to preserve the final trailing space character. That could maybe be adressed, too, if looked at on the whole.~~
This proposed fix also moves the space between the colon and the URL to the mustache template.

Best,
Luca


timhunt:
Thanks Luca :-)

Issue #376, opened 2022-01-25
Language localization in status notifications to students


cdipe:
(I'm not sure if this has been solved in later versions, we use 4.4.2)
In the StudentQuiz Digest Notification to students the status word is in English, the rest of the text is localized.

Sample in Swedish:
"Din fråga Vilka djur gör vad? i StudentQuizet StudentQuiz: Att arbeta med IKT och lärande vid Lnu i kursen Demo StudentQuiz har blivit **disapproved** av Anonym ansvarig, 25 januari 2022, 10:56."

Also, there is a (minor but still) language problem concerning alla singular and plural statuses. 
For instance, "Approved", "Disapproved" etc are the same in singular and plural when in English, but in Swedish (my language) one question is "Godkänd" (approved) and multiple questions are "Godkända". This concerns all the statuses: When filtering questions the plural word is needed, but when setting the status and in the notifications there should be the singular word.

timhunt:
Is this the same as https://github.com/studentquiz/moodle-mod_studentquiz/issues/373 ? 

(The fix for that was just integrated, but not released yet.)

cdipe:
@timhunt I'm not sure. 
But the only problem I found in the message was the status word, "Approved", "Disapproved" etc. The rest of the message was in correct Swedish.

lucaboesch:
I can confirm, @timhunt, that this is not #373.
@cdipe note that you can improve this yourself. The string state_approved | studentquiz is not translated in sv.
You can contribute to the translation by using the https://lang.moodle.org page or translate this in your own Moodle instance (Sharing it with the whole swedish speaking community by contributing this in https://lang.moodle.org would be lovely though).

You can close that one, Tim.

lucaboesch:
Ah. I see. It is about creating a appovedplural and approvedsingular / disappovedplural and disapprovedsingular string.
That doesn't change the fact that tate_approved | studentquiz is not translated in sv is not translated in AMOS (https://lang.moodle.org).

cdipe:
Hi @lucaboesch, I have translated the whole package in my installation (except the privacy statements, which we don't). But it seems that state_approved and the other statuses don't end up in the notifications.
![image](https://user-images.githubusercontent.com/14754240/150964366-a972699b-73ca-4b44-8140-c94b67c47e6f.png)
I intend to upload the language pack when I'm done.

cdipe:
[...]

PR #377, opened 2022-01-25
Past participle, past participle plural and notification localization.

lucaboesch:


lucaboesch:
I think that should be it.
Probably this addresses everything in #376.

Here are the strings now given out:
grafik

grafik

I have no mailing testing system (like mailhog) here but I think this should be fine for the 'emailsinglebody' mailing message as well which now gets a localized string.

Best,
Luca

timhunt:
Thank you very much, Luca, for correcting our English 😉.

But, seriously, this is important polish. Thank you for helping to make StudentQuiz better.

PR #379, opened 2022-02-11
Fix bug missing some override permissions of context module

nguyenphuctien:
Some of permissions we set in module (activity) level are removed by the ensure_relation script.
I think that the "no meaning" in the comment is not correct. 

            _// After going through, all remaining caps are excessive have to be usassigned. If there are capabilities in
            // the list not related to required or override, they have **no meaning** anyway, since this list only contains
            // unresolved capabilities._
            
Some of new SQ capabilities allow to override in activity level, but they do not in the list of $studentquizrelation, such as mod/studentquiz:canselfratecomment, mod/studentquiz:pinquestion, mod/studentquiz:canselfcommentprivately, etc.

Steps to reproduce:
1. Remove a role (such as tutor in my video) from the roles with permission list of capability mod/studentquiz:canselfratecomment in context course
2. Add the role about to roles with permission list of capability mod/studentquiz:canselfratecomment in context activity
3. Login in other browser with a user with the role above and access to the SQ above.
4. Reload the allowed list of capability mod/studentquiz:canselfratecomment in the step 2.
Expected result: The selected role (tutor) should be shown in the Roles with permission list of capability mod/studentquiz:canselfratecomment
Actual: The selected role disappear after we reload page.

https://user-images.githubusercontent.com/1708403/153570273-2938dc66-b05f-4a2a-a738-45a4824e95a7.mov

I suggest to remove the code which remove our override capabilities or we add the new capabilities to the $studentquizrelation.

Could you please review it and give me some advice?

### DIFF
--- a/classes/access/context_override.php
+++ b/classes/access/context_override.php
@@ -136,7 +136,7 @@ class context_override {
      * outcome of this chain of events may be uncontrollable and thus should be avoided or filtered very carefully!
      *
      * @param context $context where to apply the overrides.
-     * @param array $relation where keys are needed capabilities and its values an array of capabilities to override
+     * @param array $relation where keys are StudentQuiz capabilities and its values are array of needed capabilities.
      */
     private static function ensure_relation(context $context, array $relation) {
         global $DB;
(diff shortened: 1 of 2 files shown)


nguyenphuctien:
Sorry I just realise that the removed code is needed for the case we remove the SQ capability, the phpunit test said that. I will try to fix it in other way.

nguyenphuctien:
@dionysius
It's not about the role assignment. That screen shows the roles which allow the capability in activity level.

In the ensure_relation function, the $overridecapnames contains all custom capability for the sq in activity level (not only the question's cap, but also all other caps such as sq's caps, course's caps, system's caps, which have been modified in the activity level (if the capability has 'contextlevel' = CONTEXT_MODULE, we can do it).
In my case, the $overridecapnames contains the mod/studentquiz:canselfratecomment when we do the check for the tutor role. It will be unassigned at the end, so the tutor will be disappear from the list of mod/studentquiz:canselfratecomment in activity level.

I have fixed the code to just remove the needed capabilities in the $studentquizrelation instead of remove all the capabilities in the overridecapnames. I hope that it's correct with the old logic.

dionysius:
I see. I'm just surprised this bug took until now, as every stundentquiz capability not listed in the array would have had this issue.

Looks good to me so far if you agree that this way now every override is not getting removed if it's not in that array.

timhunt:
Arght! what we have here is some complicated logic, with some random changes, which allegedly fixes some bug.

What we do not have here are any automated tests, which clearly show what the expected behaviour is.

There is one existing test, in test_context_override, but clearly that was not good enough to find this bug. Also, it is pretty obscure what that is actually testing for.

If you have something complex to test, with tricky setup, it is worth investing in the test code to make it as readable as possible. (For example https://github.com/moodle/moodle/blob/e94167174a552e367b68a9e50188cd5181d00ade/mod/quiz/tests/structure_test.php#L241). And, this is not the first time we have had bugs in this code. So, let's get the tests right now, please.

timhunt:
Also, if we need to look up values in arrays, then having to in_array and array_search just makes the code unnessarily hard to read. Instead of doing

```php
$resolvedcapnames[] = $cap;
```

we can instead choose to do 

```php
$resolvedcapnames[$cap] = CAP_ALLOW;
```

which almost certainly leads to the rest of the code becoming simpler.

The other thing that would make this code much more readable would to to fix the variable names. What is a 'resolved' capability? I don't know.

The fact that this code only consideres CAP_ALLOW is almost certainly wrong.

nguyenphuctien:
Thank @timhunt,

I will update the code and add some more automated tests to cover the logic.

nguyenphuctien:
Hi @timhunt ,

I have updated the code. Could you please review it again?

timhunt:
Sorry, that was completely wrong - and your automated tests were clearly insufficient to detect that.

I think you have changed the code back to how it was in the past, at a time when it was even more buggy than it is now.

I think what is required is this:
https://github.com/timhunt/moodle-mod_studentquiz/commit/467a89c487788e7ecfeb3bf0d9fe37fed6ee98c1
Acutally, that Diff is essentially unreadable.

Easier to just look at the new code: https://github.com/timhunt/moodle-mod_studentquiz/blob/467a89c487788e7ecfeb3bf0d9fe37fed6ee98c1/classes/access/context_override.php#L138 and the new tests: https://github.com/timhunt/moodle-mod_studentquiz/blob/467a89c487788e7ecfeb3bf0d9fe37fed6ee98c1/tests/permissions_test.php#L41

I am going to look at this again in the morning, once I have slept. (For example, I will probably add some more test cases.)

@nguyenphuctien (any anyone else) please can you look too, and see if you can spot any case I have failed to consider. Also is it possible for anyone other than me to understand what this code is doing and why? I did my best to make it clear, but that was not easy.

(And example of how tricky it was. Until very near the end, the 'non-editing teacher role set back to default after capabilites were assigned in the past' test case was failing. That is why the 'After doing that it is important to re-fetch the current permissions.' bit was added, and some steps in the process re-ordered.)

nguyenphuctien:
Hi @timhunt ,

The code in https://github.com/timhunt/moodle-mod_studentquiz/commit/075997a35743589db1f45cb8d3f93be7e7f3e1e6#diff-7ce0a1c1ce9cad4274da94d5d3b04542513c86e5d57e08f7a3ea4528926280b3R179 fixs this bug, and the other comments and test cases make me understand clearly how it works.

timhunt:
Right, I have opened pull #387 with my changes, so I am closign this one now.

PR #383, opened 2022-03-02
Fix git action grunt error

nguyenphuctien:
Hi @timhunt ,
Moodle has updated grunt, so the git action throws errors.
I have fixed the errors in this commit. Could you please help me review it?


nguyenphuctien:
I have updated the latest version of error files, but it's not working as expected. I will checking them again. Sorry for that, I was careless!

timhunt:
Thanks Tien. Merged.

PR #385, opened 2022-03-04
Display error page when Tutor does have 'change state' capability

JBThong:
Hi @timhunt,

In this commit, I have fixed some UI to improve UX.
When the tutor does not have 'changestate' permission set. The page throw error as the image below:
image

I have listed cases I have changed when the tutor does not have 'changestate' permission
1. 'Change state' button will be hidden  in the studentquiz view
image

2. The icon in state column will be disabled
image

3. To ensure security, when we copy the URL change state of questions from browser A to another browser, 
I have displayed a warning message.
image

I have also updated behat test for this case.
Many thanks.


timhunt:
Thanks. That looks good. Merged.

PR #388, opened 2022-03-15
Fix filter errors and add behat in the advance filter

JBThong:
Hi Tim, 

In this commit, I have fixed errors in the filter of the Studentquiz as the image below:
telegram-cloud-photo-size-5-6160990514799817129-x

I. I have update the new form validation in client side:
image

image

II. I have also fixed some bugs relating to the filter:
 1. Enable Creation date is after/is before throws error
    - At creation we check enable date is after/is before then click on 'Filter' button. The filter can be run without errors message
 2. Fix validation message display in the responsive UI. 
image

III. The git action errors are not correct:
1. The function names use to override methods from the base class, so I can not update their names.
2. Do not require phpdoc for methods that override base class.

Many thanks, Tim

### DIFF
--- a/classes/question/bank/question_bank_filter.php
+++ b/classes/question/bank/question_bank_filter.php
@@ -163,7 +163,7 @@ class studentquiz_user_filter_date extends user_filter_date {
                     }
                 }
                 if (!empty($dateselector)) {
-                    $isbefore = optional_param('timecreated_sdt', 0, PARAM_INT);
+                    $isbefore = optional_param_array('timecreated_sdt', 0, PARAM_INT);
                     if ($isbefore && $isbefore['enabled']) {
                         // The first active element is "Day" selection.
                         $targetelement = $dateselector[0];
@@ -423,19 +423,19 @@ class user_filter_number extends studentquiz_user_filter_text {
         // When a count doesn't find anything, it will return NULL, so we have to account for that.
         switch($operator) {
             case 0: // Higher.
-                $res = "$field > :$name1 OR ($field IS NULL AND 0 > :$name2)";
+                $res = "$field > :$name1 OR ($field IS NULL AND 0.0 > :$name2)";
                 $params[$name1] = $value;
                 $params[$name2] = $value;
                 break;
             case 1: // Lower.
-                $res = "$field < :$name1 OR ($field IS NULL AND 0 < :$name2)";
+                $res = "$field < :$name1 OR ($field IS NULL AND 0.0 < :$name2)";
                 $params[$name1] = $value;
                 $params[$name2] = $value;
                 break;
             case 2: // Equal to.
                 $res = $DB->sql_equal($field, ":$name1", true, false)
                         . " OR ($field IS NULL AND " .
-                        $DB->sql_equal("0", ":$name2", true, false) . ")";
+                        $DB->sql_equal("0.0", ":$name2", true, false) . ")";
                 $params[$name1] = floatval($value);
                 $params[$name2] = floatval($value);
                 break;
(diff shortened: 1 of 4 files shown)


tests/behat/filter.feature
```diff
@@ -0,0 +1,76 @@
...
+    And the following "questions" exist:
+      | questioncategory               | qtype     | name            | questiontext          |
+      | Default for StudentQuiz Test 1 | truefalse | Test question 1 | Answer the question 1 |
+      | Default for StudentQuiz Test 1 | truefalse | Test question 2 | Answer the question 2 |
+
+  @javascript @_switch_window
+  Scenario: Check validation client in the advance filter.
+    When I am on the "StudentQuiz Test 1" "mod_studentquiz > View" page logged in as "admin"
+    And I should see "Show more..."
+    Then I click on "Show more..." "link"
```
timhunt, line 29: The 'Then' in a Behat test is meant ot indicate when some expected thing is being verified. That is defitely not what is happening here.

Possibly, the best place for it in this Scenario is line 75 '    And I should see "Test question 1"' or line 31 '    And I should see "You must enter a number here."'

Also, no benefit in checking 'I should see "Show more..."' just before you do 'I click on "Show more..." "link"'. The 'click' check will also test that it is visible, so this is redundant.

tests/behat/filter.feature
```diff
@@ -0,0 +1,76 @@
...
+    And I should see "You must enter a number that is greater than 0."
+    And I set the field "My difficulty value" to ""
+    And I should not see "You must enter a number that is greater than 0."
+
+    And I set the field "My Rating value" to "TF 01"
+    And I should see "You must enter a number here."
+    And I set the field "My Rating value" to "-2"
+    And I should see "You must enter a number that is greater than 0."
+    And I set the field "My Rating value" to ""
+    And I should not see "You must enter a number that is greater than 0."
```
timhunt, line 70: I have to say, if it was me, I would find a way to check all this detailed validation logic in PHPunit (faster!) and then all the Behat test has to verify is that one typical validation message appears in the UI. However, you have coded this now, so you could leave it as is.

classes/question/bank/question_bank_filter.php
```diff
@@ -458,12 +479,12 @@ class user_filter_percent extends user_filter_number {
...
      */
     public function get_sql_filter($data) {
-        $val = round($data->value, 0);
+        $val = round($data['value'], 0);
         if ($val > 100 or $val < 0) {
             return '';
         }
         if ($val > 1) {
-            $data->value = $val / 100;
+            $data['value'] = $val / 100;
```
timhunt, line 487: If $data is an array, please update the PHPdoc to say that.

(Although, acutally, this is an overridden method, so the PHPdoc comment should be deleted.)

classes/question/bank/question_bank_filter.php
```diff
@@ -444,6 +444,27 @@ public function get_sql_filter($data) {
...
+
+        if ((!is_numeric($data['value']) && !empty($data['value'])) ||
+                $data['value'] < 0 || $data['value'] > PHP_INT_MAX) {
+            return false;
+        }
+
+        return $data;
+    }
+
+    public function setupForm(&$mform) {
```
timhunt, line 459: It has not been necessary to pass objects be reference in PHP since PHP 5.0. Can we delete the '&' please. (Applies in several places throughout the patch.)

timhunt: Also, this is causing CodeChecker to fail. You have to use an invalid method name because of the parent class, but please tell CodeChecker that: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file

classes/question/bank/question_bank_filter.php
timhunt: Except that, it seems that in StudentQuiz, the consistently document overridden methods, to keep moodle-plugin-ci passing. So stick to that policy and keep the comment and fix it.

classes/question/bank/question_bank_filter.php
```diff
@@ -444,6 +444,27 @@ public function get_sql_filter($data) {
         }
         return array($res, $params);
     }
+
```
timhunt, line 447: I know that the general policy is to not comment overridden methods. However, it seems that in StudentQuiz, they consistently document overridden methods, to keep moodle-plugin-ci passing. So, we should probably follow that.

timhunt (changes requested):
In addition the  the specific comments here, the CI build is failing.

Please can you allow the CI build to run to completion on your github, and verify it is green, before creating the pull request. Thank you.

JBThong:
Hi @timhunt, 

I have updated my code as  your suggestion in this [commit](https://github.com/JBThong/moodle-mod_studentquiz/commit/3264596a7bbaaa466432981e0a215a969ad16a63) :

1. Update behat: Remove the redundant step.
2. Update PHP docs for overridden methods.
3. I have removed "positiveint" rule in the form group because its support for moodle version is greater than 3.10. Which has mentioned in the ticket [MDL-71126](https://tracker.moodle.org/browse/MDL-71126)
4. About the method setupForm(&$mform), you have suggested to me remove the '&' but this is an overridden method so I think we should keep it to prevent system warning:
"Declaration of user_filter_number::setupForm($mform) should be compatible with studentquiz_user_filter_text::setupForm(&$mform)".
Do you agree with this point?

Could you please help me to review it again, thanks?

timhunt:
Thanks. That looks good now. Merging.

PR #389, opened 2022-04-15
Studentquiz: Update css for displaying error form-group

JBThong:
Hi Tim,

I have created the commit to change the approach CSS in form-group filter.
As I have discussed with you, we have no choices in the situation so I have set CSS important but it makes the cibot warning. I think we ignore the cibot warning in this case.

Thanks

image

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -822,15 +822,15 @@
 .path-mod-studentquiz #categoryquestions .header.state {
     text-align: center;
 }
-/* CSS style for message validation form, because the align-items-center is set important in the parent of form-group. */
-@media only screen and (min-width: 396px) {
-    .path-mod-studentquiz #form-advanced-div .form-control-feedback.invalid-feedback {
-        position: absolute;
-    }
-}
 
-@media only screen and (min-width: 576px) {
-    .path-mod-studentquiz #form-advanced-div .form-control-feedback.invalid-feedback {
-        margin-top: 4.5em;
-    }
+/* Specific CSS override align-items-center, we need to prevent align-items: center in the form-group. */
+.path-mod-studentquiz #fgroup_id_rate_grp div.align-items-center,
+.path-mod-studentquiz #fgroup_id_rate_grp div.align-items-center,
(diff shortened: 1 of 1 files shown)


dionysius:
Oh, I could have a look at that as well, but easter is around the corner and I have no development setup ready.

Rule of thumb: the most specific css path wins, so if we add a better path we might get rid of the !important part. But maybe you've already discussed that.

Although this task is impossible if the previous style has already !important, which might be the case.

JBThong:
Yes, thanks @dionysius,

The align-items-center class has !important CSS. I have discussed with Tim about this. We will back to it after the easter.

Happy Easter!! :))

timhunt:
If we are going to do !important, then we need to ensure it does not break CI (https://stylelint.io/user-guide/ignore-code/)

However, do we really need to override this? What happens if we just leave the standard Boostrap CSS here?

styles.css
```diff
@@ -822,15 +822,14 @@
...
-        margin-top: 4.5em;
-    }
+/* Specific CSS override align-items-center, we need to prevent align-items: center in the form-group. */
+#fgroup_id_rate_grp div.align-items-center,
+#fgroup_id_rate_grp div.align-items-center,
+#fgroup_id_difficultylevel_grp div.align-items-center,
+#fgroup_id_publiccomment_grp div.align-items-center,
+#fgroup_id_myrate_grp div.align-items-center,
+#fgroup_id_myattempts_grp div.align-items-center,
+#fgroup_id_mydifficulty_grp div.align-items-center {
```
timhunt, line 833: We must make sure this style rule does not break layout in other plugins.

Each selector should start with .path-mod-studentquiz or similar (assuming that we really cannot avoid having this.)

timhunt (changes requested):

JBThong:
Thanks for your feedback, Tim

I have added .path-mod-studentquiz and updated the style lint to ignore the important!

Could you please help me to review it again?
Thanks,

timhunt:
Thanks Thong. Merged.

PR #390, opened 2022-04-19
Update to adapt Moodle 4.0

vuvanhieu143:
Hi @timhunt ,
Could you please help me to peer-review this PR:
This PR including code changes to make StudentQuiz work with Moodle 4.0
+ Implement new logic to support question versions.
+ Privacy API (GDPR)
+ Database structure changes(Install and upgrade)
+ Update Unit tests and behat tests.
+ Update backup/restore process
+ Update Import/export question.

Many thanks.


timhunt (changes requested):
First of all, I want to say that there is a huge amount of excellent work here. Well done.

These comments are based on the patch up to the start of renderer.php. I will review the rest tomorrow.

A) I have made a lot of comments but most of them are minor, and I hope won't take too long to fix up. Some don't even require you to make a change (as part of this work), they are just things I noticed on the code that I wanted to comment on.

B) I think the only bit (that I have seen so far) that requires significant work is the upgrade code. See the detailed comments that I made there.

C) And, there is one thing I don't see which I think should be added. In tests/fixtures, there are various example backup files, with Behat tests to check that they restore correctly. (And, of course, the existing tests with the existing files are a useful test.) However, I think we should add at least two more example backup files, one made with the previous version of StudentQuiz running in 3.11, and one make with the new StudentQuiz for Moodle 4.0. Those files should have a small amount to typical data, to test as much as possible of the backup/restore code, including restoring a 3.11 backup into 4.0.

D) As you work on the remaining fixes, please, please, please can do you all the changes as separate commits on your branch, to make it easier to re-review just the changes. Once we have finished reviewing and changing everything, then I will do a final rebase to squash it all together. Thanks.

But, as I say, this is basically a lot of good work. We just need to finish off and tidy up the last few bits.

franzvinzenz:
the file db/upgrade.php generate SQL-Error in mySQL/MariaDB.
the attached File should work without errors.

it generates an SQL-Error:
mod_studentquiz
2022052300.01: Success (0.27 seconds)
2022052300.02: Success (0.04 seconds)
2022060500.01: Success (0.22 seconds)
2022060500.02: Success (0.03 seconds)
Error writing to database
[More information about this error](https://docs.moodle.org/400/en/error/moodle/dmlwriteexception)
×Debug info: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'FROM (SELECT sq.id as studentquizid, q.id as questionid
' at line 3
UPDATE mdl_studentquiz_question sqq1
SET studentquizid = sqd.studentquizid
FROM (SELECT sq.id as studentquizid, q.id as questionid
FROM mdl_question q
JOIN mdl_question_versions qv ON q.id = qv.questionid
JOIN mdl_question_bank_entries qbe ON qv.questionbankentryid = qbe.id
JOIN mdl_question_categories qc ON qc.id = qbe.questioncategoryid
JOIN mdl_context ctx ON ctx.id = qc.contextid
JOIN mdl_studentquiz sq ON sq.coursemodule = ctx.instanceid
JOIN mdl_studentquiz_question sqq ON sqq.studentquizid = q.id
) as sqd
WHERE sqq1.studentquizid = sqd.questionid
[array (
)]
Error code: dmlwriteexception

[upgrade.zip](https://github.com/studentquiz/moodle-mod_studentquiz/files/9250721/upgrade.zip)

lucaboesch:
Once merged you might also want then to modify the README.md file.

vuvanhieu143:
Hi Tim,
I have updated the code base on your previous feedback and including the recent changes in StudentQuiz as well.
so all the new behavior for StudentQuiz should working well with 4.0.

classes/question/bank/studentquiz_bank_view.php
```diff
@@ -676,9 +677,9 @@ protected function parse_subsort($sort): array {
     /**
      *  Return the all the required column for the view.
      *
-     * @return \question_bank_column_base[]
+     * @return array \question_bank_column_base[]
```
timhunt, line 680: Acutally, 

`@return \question_bank_column_base[]`

is valid PHPdoc, and gives more information than just array, so better to use that.

classes/utils.php
```diff
@@ -622,13 +582,13 @@ public static function update_studentquiz_progress(\stdClass $studentquizprogres
...
-    public static function question_save_action(int $studentquizquestionid, int $userid = null, int $state,
+    public static function question_save_action(int $studentquizquestionid, ?int $userid, int $state,
             int $timecreated = null) {
-        global $DB, $USER;
+        global $DB;
 
         $data = new \stdClass();
         $data->studentquizquestionid = $studentquizquestionid;
-        $data->userid = isset($userid) ? $userid : $USER->id;
+        $data->userid = $userid;
```
timhunt, line 591: Is this change really safe? The funciton still seems to allow userid = null to be passed in, so don't we need this?

timhunt: Oh, I see. The DB scheme was changed to allow null. OK.

tests/behat/commen_history_backup_restore.feature
```diff
@@ -12,7 +12,8 @@ Feature: Restore of studentquizzes in moodle exports with comment histories
   @javascript @_file_upload @_switch_window
   Scenario: Restore moodle backups containing history comments.
     Given I am on the "C1" "Course" page logged in as "admin"
-    And I navigate to "Restore" in current page administration
+    And I navigate to "Course reuse" in current page administration
+    And I select "Restore" from the "jump" singleselect
     And I press "Manage backup files"
     And I upload "mod/studentquiz/tests/fixtures/backup-moodle311-c1-historycomment.mbz" file to "Files" filemanager
```
timhunt, line 18: Rather than all these many steps of Behat code, can't we use one of the standard steps like

`And I restore "test_backup.mbz" backup into "Course 2" course using this options:`

I am not sure exactly which of the steps from backup/util/ui/tests/behat/behat_backup.php does what we need, but hopefully there is one.

timhunt: Also, please can we rename this file, so that 'comment' is spelt correctly ;-)

timhunt:
Thanks for continuing to work on this. I seems like you have done a good job of fixing everything from the first review.

Having revewed the changes since the first patch, I only have a few more questions, which I added as comments in the code.

timhunt (changes requested):

version.php
```diff
@@ -28,7 +28,7 @@
 defined('MOODLE_INTERNAL') || die();
 
 $plugin->component    = 'mod_studentquiz';
-$plugin->version      = 2022020200.02;
+$plugin->version      = 2022080300.08;
```
timhunt, line 31: Note, Moodle plugin database does not like decimal points in plugin version numbers. Please change this to be

2022080308 instead of 2022080300.08

that is what those last two digits of the vertsion number are for.

(The use of decimal points in Moodle core is for moodle core top-level version.php only.)

tests/behat/commen_history_backup_restore.feature
vuvanhieu143: Hi Tim, I will rename the files.
regarding the behat step. I had tried to find all the step in backup/util/ui/tests/behat/behat_backup.php
You can see that use this step in same scenario as well: https://github.com/studentquiz/moodle-mod_studentquiz/blob/bca9dc6b33e6128edaf84d8f8055217719ccfe30/tests/behat/commen_history_backup_restore.feature#L20
Unfortunately, we don't really have shorter step in the boost theme to use.
Because the step above need to be in the restore page, as the doc said: "You should be in the 'Restore' page where the backup is." and to go to the restore page in boost theme, we need those two step above.
Is it ok that we keep those steps?

vuvanhieu143: Hi Tim, just ignore my comment.
I found the steps to do this thank for your sharing on the behat improvement on the SIG meeting.
I'm update the behat test now.

timhunt (approved):

timhunt:
Right, I think we got there. Thank you very much Hieu.

I merged your change while making the following changes:

# I removed the unwanted [Attempt to write a script that automatically tests upgrade](https://github.com/studentquiz/moodle-mod_studentquiz/pull/390/commits/35031bb0118a57355d4d17b2ed81891cd42ff61b) commit.
# I added some additional comments in db/upgrade.php to make it clearer what was going on.
# I am about to update the readme files etc.

PR #395, opened 2022-05-17
Studentquiz: Missing the History of comment when Backup - Restore to …

JBThong:
Hi Tim, 

On this commit, I have fixed bugs missing the history comment when we backup-restore, I also add the new behat to cover it.

Thanks,


timhunt:
This looks like a good fix. Merging.

It is sad that you could not avoid the use of css_element in the Behat test.

(At some time in the future, we should use https://docs.moodle.org/dev/Writing_acceptance_tests#Custom_selectors_.28..._in_the_.22....22_.22....22.29 so that instead we can say `And I should see "Deleted post" in the "2" "mod_studentquiz > question_comment"` (or something like that). But, that can be done in a future change.)

PR #396, opened 2022-05-20
Fix conflict when restore a backup from different site.

vuvanhieu143:
- Fix the issue "include admin user" in the backup file causing conflict when restore on a different site.
- Set the userid in studentquiz_state_history table to nullable.
- Fix studentquiz_delete_instance and add unit test for it. 

Could you please help me to peer-review it?

Many thanks.


timhunt:
Thanks Hieu, this looks very good. I just spotted a few minor issues.

1) In the behat tests, `And I should see "" in the "Question visibility set to 'Shown'" "table_row"` is a meaningless test. (well, it is probably about the same as `And the "Question visibility set to 'Shown'" "table_row" should exist'

Actually, I think that when there is no user, if would be better to display '-' instead of ''. Please can  you make that change, and the update the test to look for that.

2) In version.php would be better to use today's date, rater than 2022051804.02.

3) Well done for finding and fixing the missing code studentquiz_delete_instance. Ideally, this fix would be done in a separate commit, with a clear commit comment to explain what is being done. Can you do the necessary git fiddling to split this change into two separate commits?

timhunt (changes requested):
As in the 'Coversation' tab.

vuvanhieu143:
Hi @timhunt , I have updated the code and separated into two commits. Could you please peer-review again?
Many thanks

timhunt (approved):

timhunt:
Thanks. changes merged now.

PR #398, opened 2022-05-25
Fix missing pagination after create/edit question.

vuvanhieu143:
Hi Tim,
I have added the missing qperpage param when we edit/create a question.
Could you please help to peer-review it?

Many thanks


timhunt:
See my comment in the OU bug tracker, about why this very specific fix to this very specific bug report is not the right approach. Closing.

PR #406, opened 2022-06-09
StudentQuiz: Fix shorten text in comment to display correctly.

vuvanhieu143:
Hi Tim,
I have removed the custom shorten text in SQ and using the core function.

Could you please help me to peer-review again?

### DIFF
--- a/classes/commentarea/comment.php
+++ b/classes/commentarea/comment.php
@@ -30,7 +30,7 @@ use popup_action;
 class comment {
 
     /** @var int - Shorten text with maximum length. */
-    const SHORTEN_LENGTH = 160;
+    const SHORTEN_LENGTH = 75;
 
     /** @var string - Allowable tags when shorten text. */
     const ALLOWABLE_TAGS = '<img>';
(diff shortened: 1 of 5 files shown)


timhunt:
1. Moodle code should not use htmlspecialchars. It should use Moodle's equivalent, which is s(). However, should we even be doing this? Why are we not doing the escaping when the text is put into the HTML? That would be more normally, and avoids double-escaping bugs. Anyway, the code is not self-evidently right, so it would be nice to see the automates tests covering this.
2. (The code would be easier to review if anything about the convert_to_object function (name, PHPdoc comment) explained what this function was acutally doing.)
3. When calling funcitons, don't pass final optional arguments that match the default. (e.g. , false, '...' for shorten_text). If you are using PHP storm, it warns you about this.

vuvanhieu143:
1) Moodle code should not use htmlspecialchars. It should use Moodle's equivalent, which is s(). However, should we even be doing this? 
Hi @timhunt 
We have an edge case where the text contains an escaped HTML entity:
Example: "Test shortened text with html enity&lt;br&gt;"
After we run through the html_to_text, it will change the "&lt;br&gt;" to <br> so I need to s() to escape it again.
"Why are we not doing the escaping when the text is put into the HTML?"
The shorten text should only contains the plain text version of our full text. I don't think we need to escape this, but I found we didn't format_text when we output to template,so I add it to the code.Is that ok ?
2) I just add more comment to the docs.Is is make more sense now?
3) I have removed it. Sorry for the stupid question,but is there any setting I need add to my PHPStorm ? My PHPStorm did show the hint or warning at all.

PR #409, opened 2022-07-22
StudentQuiz: validate pagesize input

vuvanhieu143:
Hi Tim,
Could you please help me to peer-review the changes?
Since this is not standard Moodle form, I use the client validation from HTML input type number.
I also create a page size const for StudentQuiz so we can reuse it everywhere in SQ.


timhunt:
Thanks. Good change. Merging.

PR #417, opened 2022-09-15
Fix incorrect syntax error

vuvanhieu143:
Hi @timhunt ,
Look like I make a mistake when I fix code checker error(change or to ||).
|| does not allow to using the same variable for two side but OR allow it.
Could you please help me to review the changes?

Many thanks


timhunt:
Yes, definitely more readable to move the assignment out of the if.

(For info the issue is not with whether using the value is allowed. It is just operator precidence. So.

$x = f() or $x->property _means_ ($x = f()) or $x->property

but

$x = f() || $x->property _means_ $x = (f() || $x->property)

which is why that does not work. However, the fix you did is better than just adding brackets to the old code.

PR #421, opened 2022-10-14
Hide question status field in question bank for StudentQuiz question.

vuvanhieu143:


vuvanhieu143:
Hi @timhunt ,
This change will hide  question status field in StudentQuiz (create/edit form).
this PR will fail on 4.0 as I discussed with you.
Could you please help me to review the changes.

timhunt:
So, ideally, we need https://tracker.moodle.org/browse/MDL-75917 (backport request for https://tracker.moodle.org/browse/MDL-75844) in MOODLE_400_STABLE. However, I fear that the back-porting policy requires a three-week delay.

timhunt:
@vuvanhieu143 that commit looks like a good start, but:

1. Can we make the CSS selector a bit more specific? If there is ever another form in StudentQuiz with a 'status' field, we don't want it to be mysteriously hidden. That would end up being a major pain to debug. I suggest changing `.cm-type-studentquiz #fitem_id_status` to `body.cm-type-studentquiz.path-question-type #fitem_id_status`
2. As well as just hiding the form field, I think that we should ensure that the value is never set to draft (e.g. by someone chaning the value in the HTML before submitting the form.) So, add a check in mod_studentquiz_observer, so that if the created/edited question is somehow hidden, we immediately unhide it. (Which we might be able to do using question/bank/editquestion/classes/external/update_question_version_status.php).

Does that make sense?

vuvanhieu143:
Hi @timhunt 
1) Yes, I will do that.
2) Yes, I will make the code changes to make sure value is always set to "Ready" and write a unit test for this case.

timhunt:
Thanks. Reviewed and merged.

PR #422, opened 2022-11-02
Broken images when quiz is launched

gmanaud:
Hi,

On Moodle 3.9, added images work normally when we add them in text editor and when we preview them.
However, when we start quiz, they do not appear.

This is fixed in main branch (https://github.com/studentquiz/moodle-mod_studentquiz/commit/7af298d82a5dd3d487caed80c50039f29e5b5662), but can it be fixed in 3.x branche too ?


timhunt:
Thanks.

PR #423, opened 2022-11-07
The recipient name and the actor names are flipped in anonymous ranki…

vuvanhieu143:
Hi @timhunt ,
I have updated the old logic to make sure we are not missing the case where the user is not enrolled in the course but has manager permission in the course (for example users like admin).
I have added a unit test for this case as well.
I can't write the provider function because we need data from the setUp.

Could you please review the changes?


timhunt:
Thanks. Fix looks good. Merged.

PR #424, opened 2022-11-08
Create new renderer function for warning message

vuvanhieu143:
Hi @timhunt ,
When I try this the defect related to this 619311 on our OSEP theme.
I found out we can't style the warning message directly since they really don't have a wrapper div. We can't style it without affecting other elements.
So I add a wrapper and move it to the renderer so we can easily customize it.

Capture

Does it make sense?


timhunt:
Thanks. Fix looks good. Merged.

PR #426, opened 2022-11-15
StudentQuiz: Add module form validation rule to studentQuiz form

vuvanhieu143:
Hi @timhunt 
I have created a behat test for this case.
Because we can't identify those three select elements since they are basically the same, I need to use XPath for this case.


timhunt:
Thanks Hieu. Fix makes sense. Merging.

PR #440, opened 2023-02-07
Wip642149

vuvanhieu143:
Hi @timhunt ,

- I have fixed two issues:
+ commit 1 is adding the new default sorting for the view. 
Because we have to change the sort order, I need to update the behat to the correct order in tests/behat/pagination.feature
During testing, we found out that the pinned question is not working as expected
so I also fix that and add the behat test for that case as well.
and due to MDL-75337, we can't use the bool variable for the single button anymore, so I change it to primary so that we can pass the CI in master branch.
the reason I didn't use the constant variable is that the constant didn't exist in 4.0 and 4.1 

+ commit 2 to fix this issue.
https://github.com/studentquiz/moodle-mod_studentquiz/issues/438
and I have added a unit test for that case as well.
During testing, I found out the test test_create_rate is not making any sense, so I have updated it as well.

Could you please peer review the changes?

### DIFF
--- a/classes/question/bank/studentquiz_bank_view.php
+++ b/classes/question/bank/studentquiz_bank_view.php
@@ -224,7 +224,10 @@ class studentquiz_bank_view extends \core_question\local\bank\view {
      * Override base default sort
      */
     protected function default_sort(): array {
-        return array();
+        return [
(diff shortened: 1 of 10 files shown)


timhunt:
Thanks. Looks good. Merged.

PR #441, opened 2023-02-22
StudentQuiz: Not remember the Availability setting after Backup and R…

BruceGoodGuy:
Hi @timhunt,

I solved the following in this pull request: 
- The system doesn't remember the Availability setting after Backup and Restore the student quiz activity.

During the smoke test, I discovered that the question publishing setting was also forgotten after restoring the student quiz activity. I then also fixed it on this ticket.
image

Could you please review this pull request for me? Thanks.


timhunt:
The fix is right, so thanks.

Note, it is not good to make an whole new Behat feature just to test one small bug fix. However, I created an issue so that can be solved in future, and I will merge this fix now.

PR #445, opened 2023-03-10
Wip653815

vuvanhieu143:
Hi @timhunt 

There are several commits in this branch:
+ The first three commits related to new coding rules, and node library changes in moodle, so I need to compile the js file again and upgrade the ubuntu version.

+ The fourth commit fixed the missing sesskey hidden field issue when we upload files with combined questions https://dev.azure.com/openuniversity/IT/_workitems/edit/657640
+ The last commit is to improve the backup/restore process using unit tests.
I think the time is reduced to around 6min (from 30m to 24m) when triggering the CI.

Could you please help me to peer-review the changes?


timhunt:
Thanks, that looks great, and a nice break-down into separate commits.

PR #446, opened 2023-03-28
Wip660748

vuvanhieu143:
Hi @timhunt ,
I have scanned all the actions in StudentQuiz:
This is the summary:
Only Preview, Attempt don't have sesskey check
Start quiz also doesn't have sesskey check(but they are just redirects) but still count as form submission, so I also add sesskey check. 
All actions in the filter form are extended moodle form
Hide, Change state in student quiz view, Pin already has sesskey check.
Report comment is extended from moodle form.
Create/Edit/Delete comment is using core/ajax
Change state in attempt page is using core/ajax

- The first commit is to remove unused code.
- Second is the fix for the sesskey
- Third, we need to compile a new js min/map to pass CI.
- Fourth, due to the change in MDL-77186, I need to add a private function to run the cron setup depending on the branch.


timhunt:
All the changes look OK, except the cron_setup_user but. I feel there must be a better solution to that, but I don't know what it is, so I created [MDL-77837](https://tracker.moodle.org/browse/MDL-77837). Hopefully that will give us an explanation.

timhunt:
OK, so the cron_setup_user code done here was basically right, except that since it is not a static method, it should be $this->, not self::. Also, ->cron_setup_user would be a better name. I will amend the commit to fix that up, after merging.

PR #451, opened 2023-05-25
StudentQuiz: Tutor digest notification not respecting tutor groups

vuvanhieu143:
Hi @timhunt ,

I have fixed the issue to make sure we only send to the tutors in the group that students belong to.
I add some unit tests as well.

Note: the unit test failed is not related to this issue, It will be fixed in ticket 683232

### DIFF
--- a/classes/local/studentquiz_question.php
+++ b/classes/local/studentquiz_question.php
@@ -126,6 +126,15 @@ class studentquiz_question {
     }
 
     /**
+     * Get groupid of studentuquiz question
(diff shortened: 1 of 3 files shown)


timhunt:
Oops! I clicked merge before acutally reviewing the code! Fortunately, it is good.

I did, however, improve some coment (commit 6fbb1ba858ee13634fca0f882a3a81907bd42220) based on the things I had to work out while reviewing.

PR #452, opened 2023-06-02
StudentQuiz: Deleted orphaned question is not working as expected after MDL-78025 is implemented

vuvanhieu143:
Hi @timhunt ,

This issue is ready for peer review.
I have added two separate commits:
1) To fix this issue, as I have discussed with you, we will change the status of the question to draft when we disapprove of a question in StudentQuiz.
2) To update the legacy data.

### DIFF
--- a/.github/workflows/moodle-plugin-ci.yml
+++ b/.github/workflows/moodle-plugin-ci.yml
@@ -38,6 +38,7 @@ jobs:
         include:
           - {php: '7.3', moodle-branch: MOODLE_400_STABLE, database: pgsql}
           - {php: '7.4', moodle-branch: MOODLE_401_STABLE, database: mariadb}
+          - {php: '8.1', moodle-branch: MOODLE_402_STABLE, database: pgsql}
           - {php: '8.0', moodle-branch: master, database: pgsql}
 
     steps:
--- a/classes/local/studentquiz_helper.php
+++ b/classes/local/studentquiz_helper.php
@@ -67,10 +67,16 @@ class studentquiz_helper {
     const STATE_REVIEWABLE = 7;
 
     /**
-     * Statename offers string representation for state codes. Probably only use for translation hints.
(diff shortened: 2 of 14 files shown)


timhunt:
Automated tests:

1. CI fail looks real. Needs to be fixed.
2. We should add an additional commit to update the moodle-plugin-ci config, to run on all supported Moodle branches (a bit like https://github.com/moodleou/moodle-qtype_crossword/commit/aedc498c56598968bb2abfec66f1112016517792)


Main code

1. classes/local/studentquiz_question.php - change_state_visibility - the huge if statement here is not very readable. A better way to do this, both for simplicity of code and self-documentation is: Have an array like [self::STATE_DISAPPROVED => question_version_status::QUESTION_STATUS_DRAFT] which defines the logic ...
2. Then ... well, the change_state_visibility method is not even very helpful. Why is this a separate function? I would expect the work of udpating the state of the question in the question bank to be part of the funciotn that updates the studentquiz_question state, so it is impossible for the two things to get out-of-synch. Right, we have a huge chunk of duplciate logic between changestate.php and change_question_state in externallib.php That should be fixed.
3. And, the fact that both those places end up calling change_state_visibility twice in quick succession really shows this code is not well designed.
4. classes/question/bank/studentquiz_bank_view.php build_query -  "qv.status <> '$state'", we shoudl be usign a placeholder to insert the value into the SQL.
5. ensure_studentquiz_question_status_is_always_ready - the coment here says "Make sure questions created/edited in StudentQuiz is always ready." but that is now incorrect.


Upgrade code:

1. If you are only doing a single $DB->execute, you don't need a transactoin.
2. The SQL is unnecessarily complicated. No need to have {question} and {question_versions} in the subquery (unless I am wrong). Just do UPDATE {question_versions} SET status = 'draft' WHERE questionbankentryid IN (...)
3. Then, there is no need to have two levels on nested subqueries.

timhunt:
Thanks. Nice, and speedy, work.

PR #458, opened 2023-06-29
StudentQuiz: Email notification: Display incorrect message on unhidde…

vuvanhieu143:
Hi @timhunt ,
This is a small change.
I think we can just remove the get string and pass the event directly in
https://github.com/studentquiz/moodle-mod_studentquiz/blob/main/classes/task/send_no_digest_notification_task.php#L49
but I'm unsure about that so I have updated the unit test and language text for some action notifications we are missing.
I was wondering if those language strings are necessary (maybe this is used for multiple languages)


timhunt:
That text is going into the body of an email, so it needs to be translatable.

There might be a better way to organise all these strings, to generate the right email body text, but that would be a larger change for later. For now I will merge this fix.

PR #459, opened 2023-07-31
StudentQuiz: Student can view and attempt the questions that belong t…

BruceGoodGuy:
Hi @timhunt ,

I have resolved this ticket by updating the function `require_access_to_a_relevant_group`. After spending time investigating this issue, I realized that the function `require_access_to_a_relevant_group` only covers cases where the user doesn't have the current group and tries to access the student quiz with the group mode set to separate groups.

However, in situations where the user has another group that does not belong to the valid group, this function does not handle it properly. Therefore, I believe it's necessary to update this function to address this scenario.

Regarding the use of `exit()` in this function, it makes it challenging to write unit tests for it when it always stop the code flow, so that I can't write Unite test for it.
Commit ID: [4cbcd6d8b267e83f7ff29616fa2df3a08aeb0d94](https://github.com/studentquiz/moodle-mod_studentquiz/pull/459/commits/4cbcd6d8b267e83f7ff29616fa2df3a08aeb0d94)
I'd appreciate your feedback on my changes. Please let me know your thoughts. Thank you.

### DIFF
--- a/attempt.php
+++ b/attempt.php
@@ -205,7 +205,7 @@ $navinfo->current = $slot;
 $navinfo->total = $questionscount;
 $PAGE->navbar->add(get_string('nav_question_no', 'studentquiz', $navinfo));
 
-utils::require_access_to_a_relevant_group($cm, $context);
+utils::require_access_to_a_relevant_group($cm, $context, '', $studentquizquestion);
 
 echo $OUTPUT->header();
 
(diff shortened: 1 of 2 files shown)


timhunt:
Sorry, but groups_get_user_groups is not the right function to use. We don't want course groups here, just the ones relevant to this activity, so it should be groups_get_activity_allowed_groups.

Except that, should not be implementing this logic yourself if you don't need to, and I think groups_group_visible does exactly what you need.

timhunt:
Thanks. That makes sense now. Merging.

PR #460, opened 2023-07-31
StudentQuiz: Warning message is displayed when student views the comm…

danghieu1407:
…ent history #703301

Hi @timhunt,
The issue occurred when the $instance->authorprofileurl is null. I have added the 'Null coalescing operator' to handle the case when $instance->authorprofileurl is null. This ticket is now ready for review.
Could you please review this change? Many thanks.


timhunt:
Thanks. Merged.

PR #465, opened 2023-10-02
StudentQuiz: Fix wrong param declaration

vuvanhieu143:
Hi @timhunt ,
This change is to fix the wrong declaration of sort param that causing the warning in 724262
image

So this issue is ready for review


timhunt:
So, this change alters the 'sort' argument for the mod_studentquiz_get_comments API from the old illegal value of 'false' to VALUE_REQUIRED (the default for arguments which do not specify).

I don't know what the intention was with the old, invalid code.

However, the only place this service is used passes a value for sort, so it is OK to make it required in future.

So, I will merge this.

PR #470, opened 2023-10-21
Preview window add "Close preview" button fix #469.

### LINKED ISSUE #469: The StudentQuiz question preview window has no close/return UI element
In StudentQuiz, if you open a question preview through the question bank actions the following page (studentquiz/preview.php?cmid=123&studentquizquestionid=1234&previewid=12345) has no user interface element to close the window or to return back.

lucaboesch:
Please advise whether the button should appear only in Moodle ≥ 4.0.
I hope the remaining things are correct.

Best,
Luca

### DIFF
--- a/preview.php
+++ b/preview.php
@@ -158,6 +158,12 @@ if ($question) {
 
     echo $quba->render_question($slot, $options, 'i');
 
+    $returnurl = new moodle_url('/mod/studentquiz/view.php', ['id' => $cmid]);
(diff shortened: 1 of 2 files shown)


timhunt:
StudentQuiz preview opens in a new pop-up window (or tab, as actually happens in modern browsers).

Therefore, if a close button is added, it needs to close that window/tab, not what this does.

PR #477, opened 2023-11-16
StudentQuiz: The question state automatically changes to Disapproved when no state is selected.

vuvanhieu143:
Hi @timhunt ,
I have changed the default value of the question state so we should not update the state if users haven't changed the select value.
If you want, I can write an additional javascript to disable the button if the user hasn't changed the select value but I think since this I think we shouldn't make things more complicated.


timhunt:
Shouldn't there be a Behat test for this? I will merge anyway, but consider adding one later.

PR #478, opened 2023-11-17
Incompatible with 403

ewallah:
Make sure MOODLE_403 users do not upgrade.


leumas700:
Do you know when this will be fixed? I plan on upgrading to 4.3 at the end of December.

qchan-mindfield:
Is this patch going to be fixing question bank categories parenting themselves and disappearing from front-end? This seems to be a long running issue. We've had to go into our db to correct this a couple times now.

PR #487, opened 2024-04-17
[M4.3] Default activity completion page in Couse completion setting b…

danghieu1407:
…roken #778312

Hi @timhunt,
I have added the suffixed name for the function add_completion_rules if the Moodle version is greater than  4.2.
Due to behat function being deprecated in 4.4, I have added a new behat function to support all versions.
I have updated the behat fail of studentquiz. Step "And I click on "New" "link" because in Moolde 4.4, we have a link "new section" which causes the behat to click the wrong link.
Could you please have a look at this change? Many thanks.


timhunt:
The actual fix is good.

The Behat changes are a problem. behat_course::i_add_to_section was added in Moodle 4.1 (MDL-79029). No need to define a custom Behat step.

We can drop support for Moodle 4.0 now. It has not been supported by Moodle for ages: https://moodledev.io/general/releases.

danghieu1407:
Hi @timhunt,
I have remove the 4.0. Also I have removed the custom behat and updated the behat.
Could you please have a look for this change? Many thanks

PR #491, opened 2024-05-21
StudentQuiz: The marker bullets are overlapping #787208

danghieu1407:
with the placeholder text in the comment textbox

Hi @timhunt,
This ticket is ready for review.
I have created a new function, ```setPlaceholder```, to set a placeholder when the editor is empty.
I have also added a new condition for the mutation. Previously, when we reloaded the page and then inserted a line break using "Enter", if we deleted the line break and pressed "Space", the mutation would not trigger. To address this, I added a new condition.
Additionally, I have created a regex to handle specific cases. If it matches, it means the data in the editor is empty.
Could you please review these changes? Many thanks.


timhunt:
Note for future reference: `element.dataset.placeholder = placeholder` is better than `element.attr('data-placeholder', placeholder)`.

But, I will  merge this as is, because the code is consistently using the less good style.

PR #494, opened 2024-07-15
Improve performance of statistics queries

aolley:
Without this, the many sub-queries in stats calculations that JOIN the question_references table dont hit any of the indexes on that table. For sites with large question_references tables - this can be horrible for performance as it ends up doing a seqscan on that table.

By adding the usingcontextid field, we instead get indexscan's.

### DIFF
--- a/classes/statistics_calculator.php
+++ b/classes/statistics_calculator.php
@@ -132,6 +132,7 @@ class statistics_calculator {
                       JOIN {question_references} qr ON qr.itemid = sqq.id
                            AND qr.component = 'mod_studentquiz'
                            AND qr.questionarea = 'studentquiz_question'
+                           AND qr.usingcontextid = :contextid1
                       JOIN {question_bank_entries} qbe ON qr.questionbankentryid = qbe.id
                       JOIN {question_versions} qv ON qv.questionbankentryid = qr.questionbankentryid AND qv.version = (
                                       SELECT MAX(version)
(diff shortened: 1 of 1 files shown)


timhunt:
Thank you for doing this. I should have noticed this was necessary before. Very important to hit this index. However, since this is all about query performance ...

These queries all process data from a single student quiz. Therefore, there will only every be a single contextid involved, and therefore there is no need to join the context table! Just add AND qr.usingcontextid = :contextid4 in the join on question_references, and add the extra query parameters in get_attempt_stat_joins_params. (Not my code that the params and the SQL are initialised so far apart, but now is probably not the time to fix that.)

Are you able to make this change? (If not, let me know.) Thanks.

aolley:
Good catch, you're absolutely right. I'll make that change.

aolley:
@timhunt I've updated the change to instead pass the contextid to where its needed and include it in the queries. I've done some quick testing so far and it looks like it does what it needs to do.

timhunt:
Thanks. That is better.

It is a trade-off between changing all thost function calls, and just repeating `$contextid = context_module::instance($cmid)->id` in the various functions. Contexts are cached, so repeating that is not a performance issue. Is it very annoying if I ask you to change it again?

aolley:
Well it's a function call, cached or not. Which _is_ less performant. But
hardly worth the complexity in the call changes, so sure why not. I'll
update it soon.

timhunt:
Thanks. That looks good now. Merged.

PR #499, opened 2024-08-15
StudentQuiz: improve upgrade steps with new function #814045

BruceGoodGuy:
Hi @timhunt ,

I have updated the StudentQuiz code to use a safer decode method. Could you please review it?

Thanks.


timhunt:
Thanks. Merging.

PR #501, opened 2024-08-28
StudentQuiz: Implement Fast filter UI redesign

vuvanhieu143:
Hi @timhunt, this ticket is ready for peer review.

Here’s what I’ve done in this ticket:
- Moved all the filter condition fields into separate groups so we can loop through each group. Conditions within the same group are connected with OR, while conditions in different groups are connected with AND.
- For the state_column_name string: I know it’s not consistent with other strings, but do I need to create a new string for that label?
- Added two new buttons: Mine and Answered.
- Removed some Behat tests using XPath and moved them to filter.feature. I don’t think we need to test all states like before, but  I’d appreciate your opinion. Do you think the current tests are sufficient?
- Addressed a Behat failure related to client-side validation.

Thanks!


AnupamaSarjoshi:
Thank you for working on this issue. The changes look great and have been merged.

PR #505, opened 2024-09-28
Removing margin-right in favour of Bootstrap classes.

lucaboesch:
This uses bootstrap classes instead of 5px hard coded margin.


vuvanhieu143:
Thank @lucaboesch , I forgot about this. merged.

PR #508, opened 2024-10-09
Defect 828690: Behat: mod/studentquiz (tt) - Test delete comment feat…

BruceGoodGuy:
Hi @AnupamaSarjoshi,
CC @timhunt,

I have completed the development for this ticket. The Behat failure was caused by an issue where, (maybe) due to network instability, the Tiny editor wasn't fully loaded when we tried to set the event listener, resulting in the event not being properly attached. To fix this, I’ve updated the initBindEditor method to ensure that the Tiny editor is fully loaded before proceeding with further actions.

Additionally, I’ve disabled the "Add reply" button initially, as the reply content is always empty at the start, to keep it consistent with the behavior of the "Add comment" button.

Please note, the JavaScript code in this file isn't written in a modern style, so I have not converted it to ES6. I believe this can be resolved in a separate ticket later on.

Could you please review the changes?

Thanks.


AnupamaSarjoshi:
Thanks so much @BruceGoodGuy. This looks good and is merged :)

PR #510, opened 2024-11-06
StudentQuiz: The comment/reply input field still displays the entered…

BruceGoodGuy:
Hi @AnupamaSarjoshi ,

After investigating, I found that the issue is related to the Auto Save feature in TinyEditor. Since the comment editor's ID stays the same after a user creates a comment, the draft content is automatically retrieved each time, leading to this issue.

To fix this, I used the `removeAutosaveSession` method (available in TinyEditor) to clear the draft content whenever a user successfully adds a comment.

Could you please review this update?


AnupamaSarjoshi:
Thanks @BruceGoodGuy for working on this issue. The changes look good to me and merged upstream.

